PR #15, opened 2026-05-13
feat(vault-mcp): daily note + property discovery tools (15 → 19)

aliasunder:
## Summary

- **`vault_get_daily_note`** — reads `.obsidian/daily-notes.json` for portable path resolution (folder + Moment.js date format), falls back to Obsidian defaults. Returns `{ path, content, exists }` so agents can create missing notes with the resolved path.
- **`vault_list_property_keys`** — discovers frontmatter schema via `json_each(properties)`. Returns key names, note counts, and top 3 sample values per key to reduce round-trips.
- **`vault_list_property_values`** — enumerates distinct values + counts for a given property key. Handles both scalar (`status: "active"`) and array (`tags: ["a", "b"]`) properties via `CASE json_type()` + `json_array()` wrapping.
- **`vault_search_by_property`** — property-only search without requiring a text query. Fills the gap where `vault_search` requires FTS5 input. Handles scalar equality and array containment.

New file: `src/vault-mcp/daily-notes.ts` (config reader, `momentToLuxonFormat` converter, path resolver).

All SQL uses named params (`@key`, `@value`) with two-arg `json_type(json, path)` to avoid `json_extract`-on-NULL errors in cross-joined `json_each`.

46 new tests (284 → 330).

## Test plan

- [x] `npm test` — 330 tests pass
- [x] `npm run build` — TypeScript compiles clean
- [x] Pre-commit hooks (eslint, prettier) pass
- [x] Deploy to Lightsail, verify 19 tools via Claude Desktop
- [x] Test daily note: today, specific date, missing note (check resolved path)
- [x] Test property keys: verify counts and sample values against real vault
- [x] Test property values: scalar (status) and array (tags) properties
- [x] Test property search: `key: "type", value: "session-log"` against real data

### DIFF
--- a/src/vault-mcp/__tests__/daily-notes.test.ts
+++ b/src/vault-mcp/__tests__/daily-notes.test.ts
@@ -50,6 +50,21 @@ describe("momentToLuxonFormat", () => {
       input: "YYYY_MM_DD journal",
       expected: "yyyy_MM_dd journal",
     },
+    {
+      name: "converts [literal] escapes to Luxon single-quote syntax",
+      input: "YYYY-MM-DD [Daily Note]",
+      expected: "yyyy-MM-dd 'Daily Note'",
+    },
+    {
+      name: "handles [literal] with apostrophe",
+      input: "YYYY-MM-DD [it's]",
+      expected: "yyyy-MM-dd 'it''s'",
+    },
+    {
+      name: "handles empty [literal] brackets",
+      input: "YYYY-MM-DD []",
+      expected: "yyyy-MM-dd ''",
+    },
   ]
 
   it.each(scenarios)("$name", ({ input, expected }) => {
@@ -179,6 +194,24 @@ describe("getDailyNotePath", () => {
       "invalid date",
     )
   })
+
+  it("rejects partial ISO dates (year only)", async () => {
+    await expect(getDailyNotePath(vaultDir, "2026")).rejects.toThrow(
+      "invalid date",
+    )
+  })
+
+  it("rejects partial ISO dates (year-month only)", async () => {
+    await expect(getDailyNotePath(vaultDir, "2026-05")).rejects.toThrow(
+      "invalid date",
+    )
+  })
+
+  it("rejects full ISO timestamps", async () => {
+    await expect(
+      getDailyNotePath(vaultDir, "2026-05-13T14:30:00Z"),
+    ).rejects.toThrow("invalid date")
+  })
 })
 
 // ── getDailyNote ─────────────────────────────────────────────────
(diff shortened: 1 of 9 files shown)


src/vault-mcp/search-index.ts
```diff
@@ -11,6 +11,21 @@ const isString = (value: unknown): value is string => typeof value === "string"
 
 const isDate = (value: unknown): value is Date => value instanceof Date
 
+/** Converts Date instances in frontmatter to ISO date strings (YYYY-MM-DD)
+ *  before JSON.stringify, preventing gray-matter's YAML 1.1 Date parsing
+ *  from producing full ISO timestamps in the properties column. */
+const normalizeDates = (
```
aliasunder: Can the naming of this function not be more explicit? eg `convertToIsoDateString`
`normalize..` is ok for a function that does general formatting, but this has a very specific output that we can be explicit about

aliasunder: or better `convertToIsoDateString` or `convertFrontmatterDatesToIsoDateStrings` ?

src/vault-mcp/search-index.ts
```diff
@@ -455,6 +481,167 @@ export const createSearchIndex = (dbPath: string) => {
...
+      ) element
+      WHERE typeof(element.value) IN ('text', 'integer', 'real')
+      GROUP BY element.value
+      ORDER BY count DESC
+      LIMIT 3
+    `
+    const sampleStmt = db.prepare(sampleSql)
+
+    const results: PropertyKeyInfo[] = keyRows.map((keyRow) => {
+      const bindParams: Record<string, string> = { key: keyRow.key }
```
aliasunder: what are `bindParams`?

src/vault-mcp/tool-definitions.ts
```diff
@@ -820,5 +825,175 @@ Returns: Confirmation message.`,
...
+      return safeHandler(
+        reqLogger,
+        async () =>
+          search.searchByProperty({ key, value, folder, limit }, reqLogger),
+        (results) => JSON.stringify(results.map(formatNoteMetadata)),
+      )
+    },
+  )
+
+  sessionLogger.info("registered tools", { count: 19 })
```
aliasunder, line 998: Do you think there's benefit to making type ToolName into an enum since they get converted to JS objects? Or actually creating JS object map that can be used like this

```
type Tool<T extends ToolName> = Record<Uppercase<T>, T>

const TOOL_NAMES: Tool<ToolName> = {
  VAULT_READ_NOTE: "vault_read_note"
  ...
} as const

  server.registerTool(
    TOOL_NAMES["VAULT_WRITE_NOTE"], 
    ...

sessionLogger.info("registered tools", { count: TOOL_NAMES.length })

```

src/vault-mcp/__tests__/daily-notes.test.ts
```diff
@@ -0,0 +1,277 @@
+import { describe, it, expect, beforeEach, afterEach } from "vitest"
+import { DateTime } from "luxon"
+import {
+  mkdtemp,
+  rm as removeDirectory,
```
aliasunder: this could go back to `rm` everywhere, i didn't realize it was from the Node package

src/vault-mcp/tool-definitions.ts
aliasunder: Good idea — a const object map gives us both runtime lookup AND the derived type. The clean version:

```typescript
const TOOL_NAMES = {
  VAULT_READ_NOTE: "vault_read_note",
  VAULT_WRITE_NOTE: "vault_write_note",
  // ...all 19 tools
} as const

type ToolName = (typeof TOOL_NAMES)[keyof typeof TOOL_NAMES]
```

Then `registerTool(TOOL_NAMES.VAULT_WRITE_NOTE, ...)` and `count: Object.keys(TOOL_NAMES).length`.

This would touch every `registerTool` call (all 19), so it's a cross-cutting refactor rather than something to scope to this PR's 4 new tools. I'd recommend doing it as a standalone commit after both PRs merge — clean diff, no conflict with PR #16. Want me to add it to TASKS.md?

PR #16, opened 2026-05-13
feat(vault-mcp): backlinks, outgoing links, and orphan detection (19 → 22)

aliasunder:
## Summary

- **`vault_get_backlinks`** — finds notes linking TO a given path. Core Obsidian concept — backlinks are invisible without a database query.
- **`vault_get_outgoing_links`** — finds notes a given path links TO, with `exists: true/false` for broken-link detection at the single-note level.
- **`vault_find_orphans`** — finds notes with no incoming links (disconnected from the knowledge graph). Excludes Daily Notes, Templates, and About Me by default.

New `links` table (deduplicated by source+target pair, indexed on target for fast backlink lookups). Link extraction uses fence-aware regex parsing — skips `[[wikilinks]]` and `[markdown](links.md)` inside fenced code blocks. Resolution: exact path → basename match → shortest path for ambiguous basenames.

Indexing pipeline changes:
- `upsertNote` — extracts links from body, resolves against known paths, stores in `links` table
- `removeNote` — cleans up outgoing links from deleted notes
- `rebuildFromVault` — two-pass: index all notes first, then re-extract links with the complete path list for accurate resolution

39 new tests (284 → 323).

**Note:** This PR branches from `main`. PR #15 (property + daily note tools) modifies the same files — expect trivial merge conflicts (list appends, count updates) after #15 lands.

## Test plan

- [x] `npm test` — 323 tests pass
- [x] `npm run build` — TypeScript compiles clean
- [x] Pre-commit hooks pass
- [x] Deploy to Lightsail, verify 18 tools via Claude Desktop
- [x] Test backlinks: pick a well-linked note, verify results match Obsidian's backlinks pane
- [x] Test outgoing links: verify `exists: false` for links to non-existent notes
- [x] Test orphans: verify Daily Notes/About Me excluded, known orphans appear
- [x] Test file watcher: edit a note's links, verify backlinks update without restart

### DIFF
--- a/src/vault-mcp/search-index.ts
+++ b/src/vault-mcp/search-index.ts
@@ -89,19 +89,19 @@ type NoteRow = {
 
 // ── Link extraction ─────────────────────────────────────────────
 
-/** Matches fenced code block openers: 3+ backticks or 3+ tildes. */
-const FENCE_OPEN = /^(`{3,}|~{3,})/
+/** Matches fenced code block openers: 0-3 spaces indent + 3+ backticks or tildes (CommonMark §4.5). */
+const FENCE_OPEN = /^ {0,3}(`{3,}|~{3,})/
 
 /** Matches wikilinks: [[target]], [[target|text]], [[target#heading]],
  *  [[target#heading|text]], and embeds ![[target]]. Captures the
  *  target path/name (group 1) before any # or |. */
 const WIKILINK_RE = /!?\[\[([^\]#|]+)(?:#[^\]|]*)?(?:\|[^\]]+)?\]\]/g
 
-/** Matches markdown internal links: [text](path) or [text](path#heading).
- *  Excludes external URLs (http://, https://, mailto:) and same-page
- *  anchors (#heading). Captures the path (group 1). */
+/** Matches markdown internal links to .md files: [text](path.md) or
+ *  [text](path.md#heading). Excludes external URLs and non-.md assets
+ *  (images, PDFs). Captures the path without extension (group 1). */
 const MD_LINK_RE =
-  /\[[^\]]*\]\((?!https?:\/\/|mailto:|#)([^)#\s]+?)(?:\.md)?(?:#[^)\s]*)?\)/g
+  /\[[^\]]*\]\((?!https?:\/\/|mailto:|#)([^)#\s]+?)\.md(?:#[^)\s]*)?\)/g
 
 /** Extracts link targets from markdown content, skipping fenced code blocks.
  *  Returns deduplicated raw targets (pre-resolution). */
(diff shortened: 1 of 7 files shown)


aliasunder:
Subscribing for notifications.

src/vault-mcp/search-index.ts
```diff
@@ -150,17 +272,31 @@ export const createSearchIndex = (dbPath: string) => {
...
   // ── Index maintenance ──────────────────────────────────────────
 
   // FTS rows are managed manually (delete-then-insert) because SQLite triggers
   // combined with INSERT OR REPLACE cause FTS5 corruption.
 
-  /** Parses a note's content and frontmatter, then indexes it for search. */
+  // When a new note is indexed, any previously-unresolved link targets that
+  // match the new note's path or basename are updated to the resolved path.
+  // This handles the "link first, create later" workflow common in Obsidian.
+  const reResolveStmt = db.prepare(
```
aliasunder: so this is very specific to links - i feel like it should be closer to where it's actually used? 

Also, it's hard to grasp all the backlink logic - there's no where that gives a high level overview of what we're doing in this file

src/vault-mcp/search-index.ts
```diff
@@ -250,8 +408,27 @@ export const createSearchIndex = (dbPath: string) => {
     )
 
     db.transaction(() => {
+      // Pass 1: index all notes (content, frontmatter, FTS) — skip link
+      // extraction here; Pass 2 handles it with the complete path list.
       for (const item of items) {
-        upsertNote(item.rel, item.content, item.mtime)
```
aliasunder, line 254: can we change this to an object with named params? there's now 4 input args that need to go in order

src/vault-mcp/tool-definitions.ts
```diff
@@ -995,5 +998,141 @@ Returns: JSON array of note metadata (path, title, tags, related, folder, type,
...
+
+  server.registerTool(
+    "vault_find_orphans",
+    {
+      title: "Find Orphans",
+      description: `Find notes with no incoming links from other notes. Orphan notes are disconnected from the vault's knowledge graph — they may be forgotten or need linking from relevant notes.
+
+Example: vault_find_orphans() or vault_find_orphans({ exclude_folders: ["Daily Notes", "Templates", "About Me"] })
+
+When to use: Vault maintenance and organization. Helps identify notes that might be forgotten or need integration into the knowledge graph. Daily Notes, Templates, and About Me folders are excluded by default since those are standalone by design.
```
aliasunder, line 1104: Just want to note that these exclusions would need to be configurable somehow before i make this public as it's very specific to my Vault

aliasunder: Agreed — the hardcoded defaults work for your vault but aren't portable. The `exclude_folders` parameter already accepts a custom array, so any caller can override. For the default values specifically, they could be read from an env var or a config file (e.g. `.obsidian/vault-cortex.json`) rather than hardcoded. Added to the backlog as part of the simplify-public-deployment task.

src/vault-mcp/search-index.ts
```diff
@@ -150,6 +290,10 @@ export const createSearchIndex = (dbPath: string) => {
     `INSERT INTO notes_fts (path, title, content) VALUES (?, ?, ?)`,
   )
   const removeNotesStmt = db.prepare(`DELETE FROM notes WHERE path = ?`)
+  const deleteLinksStmt = db.prepare(`DELETE FROM links WHERE source = ?`)
```
aliasunder, line 295: same issue with these prepare statements as the other ones.

Is there any benefit to them being at the top? just curious. If there is, then let's move `reResolveStmt` back up here - we should be consistent.

Open to your ideas.

aliasunder:
Addressed the prepared statement comment: moved `reResolveStmt` back to the top with all other prepared statements. Added a comment explaining why they're all declared up front — `db.prepare()` compiles SQL once and caches it, so preparing inside a function would re-compile on every call.

src/vault-mcp/tool-definitions.ts
```diff
@@ -995,5 +998,141 @@ Returns: JSON array of note metadata (path, title, tags, related, folder, type,
...
+When to use: When you need to understand what references a note, find related context, or assess a note's connectivity. Core Obsidian concept — backlinks are invisible without a database query.
+For outgoing links (what a note links TO), use vault_get_outgoing_links. For orphan detection, use vault_find_orphans.
+
+Errors:
+- No error if the note has zero backlinks — returns an empty array.
+
+Returns: JSON with path (the queried note), backlinks (array of { path, title }), and count.`,
+      inputSchema: {
+        path: z
+          .string()
```
aliasunder, line 1024: should we ensure the string is not an empty string?

PR #130, opened 2026-06-14
feat(vault_read_note): add outline and section-scoped reads

aliasunder:
## Why

`vault_read_note` only ever returned the **entire** note body. For large notes (a long Kanban board, a long `CLAUDE.local.md`) that exceeds the tool-result token budget and truncates. The **write** side already targets sections by heading (`vault_patch_note`), so an agent could surgically *edit* `## Active` without reading `## Done` — yet couldn't *read* `## Active` without pulling the whole file. This closes that read/write asymmetry.

## What

Two new optional, mutually-exclusive params on `vault_read_note`:

- **`outline: true`** — returns the heading tree as JSON (`{ level, text, bytes }` per heading), no bodies. A cheap structure fetch: read the outline, then drill into the one section you need.
- **`heading`** (+ optional **`heading_level`**) — returns just that section: the heading line plus its body, through the next heading of the same or higher level (child headings included). Case-sensitive exact match, mirroring `vault_patch_note`'s targeting.

`properties_only` is unchanged; default (no mode) still returns the full note. The handler rejects combining more than one mode.

## How

The section-span parser (`parseHeadings` / `findHeading`) is extracted from `vault-patcher.ts` into a new shared `heading-parser.ts`, so read and patch resolve a section to the **exact same span** — "read section X" and "edit section X" line up, and there's no second parser to drift. The move is behaviour-preserving; the patcher's existing tests pass unchanged.

## Notes on semantics

- A parent heading's span (and its outline `bytes`) includes its nested child sections — consistent with what a section read of that heading returns.
- The trailing Obsidian `%%` settings block (e.g. a Kanban board's settings) is excluded from the last section, reusing the patcher's existing boundary logic.
- Section line ranges are body-relative (frontmatter excluded), matching the patcher.

## Tests

- New `heading-parser.test.ts` — direct unit tests for the extracted parser (H1–H6 spans, child inclusion, fenced-code headings ignored, trailing `%%` exclusion, not-found/ambiguous/empty errors).
- `vault-filesystem.test.ts` — `readNoteOutline` (level/text/bytes, no-headings → `[]`, frontmatter excluded, not-found) and `readNoteSection` (heading + body, child inclusion, `heading_level` disambiguation, trailing `%%` exclusion, not-found/ambiguous).
- `tool-definitions.test.ts` — handler rejects combining modes.

657 tests pass; `lint`, `build`, and `prettier:check` clean. `AGENTS.md` structure map updated to list `heading-parser.ts`; tool count unchanged (params added, no new tool).

---
_Generated by [Claude Code](https://claude.ai/code/session_019bW8ZBciNgmkKAgfFsgmzw)_

## Summary by CodeRabbit

* **New Features**
  * Added outline mode to vault_read_note tool, returning a heading structure as JSON.
  * Added heading parameter to vault_read_note for extracting specific sections by heading text.
  * Added optional heading_level parameter for disambiguating duplicate headings.

* **Tests**
  * Added comprehensive test coverage for heading parsing and section extraction.

* **Refactor**
  * Extracted shared heading parsing logic into a dedicated module.


src/vault-mcp/tool-definitions.ts
```diff
@@ -126,6 +137,27 @@ Returns: Raw markdown string (default), or JSON object of properties (when prope
...
+          ),
+        heading_level: z
+          .number()
+          .int()
+          .min(1)
+          .max(6)
+          .optional()
+          .describe(
+            "Heading level (1-6) for disambiguation when multiple headings share the same text; only applies with heading",
+          ),
```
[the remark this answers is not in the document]
aliasunder: Good catch — both fall-throughs are fixed in 789b8ce. I kept the validation in the handler rather than tightening the Zod schema (`.min(1).trim()`): `vault_read_note`'s `heading`/`heading_level` schema is intentionally identical to `vault_patch_note`'s, and a schema-level `.min(1)` here would diverge the two. Specifically:

- mutual-exclusivity now treats `heading !== undefined` as section mode, so an empty `heading` is caught instead of skipped;
- `heading_level` without `heading` returns `heading_level requires a heading`;
- dispatch is `heading !== undefined`, so `heading: ""` reaches `findHeading` and surfaces its existing `heading cannot be empty` error.

Re "Also applies to: 181-189, 207-217" — those ranges are the `outline`/`heading` branches inside this same handler, not separate tools. The only other definition with `heading`/`heading_level` is `vault_patch_note`, where an empty/absent heading is intentionally a file-level operation, so it's left as-is. Added tests for the orphan-`heading_level` and empty-`heading` cases.

---

src/vault-mcp/tool-definitions.ts
```diff
@@ -104,15 +104,26 @@ export const registerTools = (params: {
...
-Prefer vault_search when you don't know the path. Prefer vault_get_memory for ${config.memoryDir}/ files (returns content without properties).
+When to use: You know the exact path and need a specific note's content. For a large note (a long board or doc), use outline: true to see its headings, then heading: "..." to read just the one section you need — both far cheaper than pulling the whole file. Use properties_only: true when you only need properties.
+Prefer vault_search when you don't know the path. Prefer vault_get_memory for ${config.memoryDir}/ files (returns content without properties). To edit a section you've read, use vault_patch_note (same heading targeting).
 
-Returns: Raw markdown string (default), or JSON object of properties (when properties_only: true).`,
+Section boundaries: a section spans from its heading to the next heading of the same or higher level (or EOF). Child headings are included in the parent section. This matches vault_patch_note exactly, so "read section X" and "edit section X" resolve to the same span.
```
aliasunder: We don't need to specify it matches patch note exactly. This is a tool description and unnecessary tokens.

src/vault-mcp/tool-definitions.ts
```diff
@@ -126,6 +137,27 @@ Returns: Raw markdown string (default), or JSON object of properties (when prope
           .describe(
             "If true, returns parsed properties as JSON instead of full note content",
           ),
+        outline: z
```
aliasunder, line 141: Isn't there a zod mechanism to enforce that only one option can be used at a time?

src/vault-mcp/tool-definitions.ts
```diff
@@ -126,6 +137,27 @@ Returns: Raw markdown string (default), or JSON object of properties (when prope
...
+          .boolean()
+          .optional()
+          .describe(
+            "If true, returns the heading tree as JSON (level, text, and section byte size per heading) instead of any body content — a cheap structure fetch for large notes",
+          ),
+        heading: z
+          .string()
```
aliasunder, line 148: We should ensure there is min 1 char

aliasunder: Done in 3cb0284 — added `.min(1)` to the `heading` schema so an empty heading is rejected at parse time. The handler guards (`heading_level` requires `heading`, and `findHeading`'s "heading cannot be empty" for whitespace-only input) stay as defense-in-depth.

---

src/vault-mcp/tool-definitions.ts
aliasunder: Yes — Zod has `.refine`/`.superRefine`, and the SDK's `registerTool` does accept a full schema (`inputSchema?: ZodRawShapeCompat | AnySchema`), so a `z.object({...}).superRefine(...)` is possible. The catch: cross-field refinements are runtime-only — they don't serialize into the JSON Schema the MCP client/LLM sees, so "pick at most one" wouldn't be *advertised* to callers; it'd just throw at parse time, the same effective behavior as the current handler guard but with a less controllable error message. It would also make this the only one of the 23 tools registered as a wrapped `z.object().superRefine()` instead of the uniform raw-shape style (and the SDK strips the refinement during schema generation anyway). The handler check also covers the `heading_level`-requires-`heading` case in the same place. Happy to switch to `superRefine` if you'd prefer the declarative form, but it doesn't buy enforcement the handler doesn't already provide.

---

src/vault-mcp/tool-definitions.ts
```diff
@@ -134,12 +167,47 @@ Returns: Raw markdown string (default), or JSON object of properties (when prope
...
+      }
+
+      // The read modes select different content; allowing more than one would
+      // make the result ambiguous, so reject the combination up front. An empty
+      // heading still counts as section mode (heading !== undefined) so it's
+      // rejected here rather than silently falling through to a full read.
+      if (
+        [
+          properties_only === true,
+          outline === true,
```
aliasunder: Let's make this a clearly named car vs inline

src/vault-mcp/vault-operations/vault-filesystem.ts
```diff
@@ -106,6 +107,77 @@ const readNote = async (
...
+    throw new Error(`note not found: "${params.path}"`)
+  }
+  const lines = parseNote(content).content.split("\n")
+  const headings = parseHeadings(lines)
+  const outline = headings.map((heading) => ({
+    level: heading.level,
+    text: heading.text,
+    // Section span = heading line through bodyEndLine (the same span a section
+    // read returns), so the size hint matches what reading it would cost.
+    bytes: Buffer.byteLength(
```
aliasunder: Why not put this in a var and log it?

src/vault-mcp/tool-definitions.ts
```diff
@@ -149,6 +216,28 @@ Returns: Raw markdown string (default), or JSON object of properties (when prope
...
 
+      if (outline) {
+        return safeHandler(
+          reqLogger,
+          () => vaultFs.readNoteOutline({ vaultPath, path }, reqLogger),
+          (headings) => JSON.stringify(headings),
+        )
+      }
+
+      // An empty heading reaches readNoteSection and surfaces findHeading's
```
aliasunder: Don't we have a min value check now?

PR #139, opened 2026-06-16
feat: add MCP prompts (vault-orientation, memory-review, daily-review)

aliasunder:
## What

Adds the MCP **prompts** capability (`prompts/list` / `prompts/get`) alongside the existing tools. Where tools are model-driven, prompts are **user-initiated** workflows that the MCP client surfaces as slash commands, a **+** menu, or similar — their live vault content is assembled only when you run them, over the same data layer the tools use, so there's no embedded procedure that can drift, only live content plus thin, durable instruction.

Three prompts, parameterized through existing config (`MEMORY_DIR`, daily-notes settings) so they work for any vault out of the box:

| Prompt | Arguments | Purpose |
| --- | --- | --- |
| `vault-orientation` | — | Surveys folders, tags, property keys, recent notes, and the memory layer to expose the vault's conventions. |
| `memory-review` | `file?`, `max_chars?` | Reflects on the memory layer as an append-with-dates **evolution** — reads dated entries as a timeline, never "newest supersedes older" — and proposes append-only updates. Never prunes. |
| `daily-review` | `date?`, `max_chars?` | Reconciles a day's daily note against recent activity and surfaces durable facts worth saving to memory. |

The optional `max_chars` argument caps the embedded vault content per invocation, for clients with strict payload limits (default: full content).

## Notes

- **Control model:** prompts are user-controlled (per the MCP spec's tools/prompts/resources split). The model can't invoke them like tools; clients surface them as user actions. They are intentionally **not** advertised in the server `instructions` string.
- **Resilience:** every handler degrades to a valid message rather than throwing, so a prompt never hard-fails the client. `memory-review`'s `file` argument has prefix autocomplete.
- **Zero-arg correctness:** `vault-orientation` omits `argsSchema` so the SDK calls back as `(extra) =>` (an empty `{}` schema would be treated as "has args" and break parsing).
- **Logging:** each invocation emits `prompt_call` + `prompt_result` (outcome, result size, truncation flag); invalid client input logs `warn`, and unexpected failures log `error`.

## Design: memory-review is append-only

The `About Me/` memory layer is append-with-dates and meant to be read as an **evolution**: each dated entry was true when written, and the timeline read top-to-bottom is the meaning. `memory-review` narrates that trajectory and proposes only append-only updates; it explicitly does not frame a newer entry as "superseding" an older one, does not treat evolving beliefs as contradictions to reconcile, and never prunes "stale" entries. A narrow, separate correction path exists only for facts that are mis-recorded or now genuinely incorrect. This philosophy is encoded as a regression test.

## Docs

README (headline, feature bullet, Prompts section, config table), ARCHITECTURE (MCP Prompts section + Phase 1 framing), AGENTS (structure tree + prompt conventions), and `server.json` are kept in sync with the prompts surface.

## Testing

Full suite green under Node 24, including **exact, character-for-character assertions of each prompt's complete output** over deterministic fixtures, genericness via a custom `MEMORY_DIR`, every handler branch (empty / bad-input / truncation / error degradation), the completion callback, the append-only framing guard, the new logging events and their levels, and router wiring.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -252,11 +252,11 @@ Link queries use a `links` table populated from `[[wikilink]]` and `[text](path.
 
 Alongside tools, the server registers MCP **prompts** (`prompts/list` / `prompts/get`) in `prompt-definitions.ts`, mirroring the tool factory and registered per session in `mcp-router.ts`. Prompts are user-initiated (slash command / **+** menu), cost no tokens until invoked, and assemble live vault content at invocation time over the same data layer the tools use — so there is no embedded procedure that can drift, only live content plus thin, durable instruction.
 
-| Prompt              | Arguments         | Purpose                                                                                                |
-| ------------------- | ----------------- | ------------------------------------------------------------------------------------------------------ |
-| `vault-orientation` | —                 | Surveys folders, tags, property keys, recent notes, and the memory layer to expose vault conventions.  |
-| `memory-review`     | `file` (optional) | Reflects on the memory layer as an append-with-dates **evolution**; proposes append-only updates only. |
-| `daily-review`      | `date` (optional) | Reconciles a day's daily note against recent activity and feeds durable facts back into memory.        |
+| Prompt              | Arguments             | Purpose                                                                                                |
+| ------------------- | --------------------- | ------------------------------------------------------------------------------------------------------ |
+| `vault-orientation` | —                     | Surveys folders, tags, property keys, recent notes, and the memory layer to expose vault conventions.  |
+| `memory-review`     | `file?`, `max_chars?` | Reflects on the memory layer as an append-with-dates **evolution**; proposes append-only updates only. |
+| `daily-review`      | `date?`, `max_chars?` | Reconciles a day's daily note against recent activity and feeds durable facts back into memory.        |
 
 Each handler degrades to a valid message rather than throwing, so a prompt never hard-fails the client. `memory-review` is deliberately append-only: it reads dated entries as a timeline (each entry true when written), never as "newest supersedes older," and never prunes "stale" entries — matching the memory layer's design.
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -24,8 +24,10 @@ API Gateway, SST — but Vault Cortex runs anywhere Docker does.
 ## Phasing
 
 **Phase 1** delivers vault CRUD, full-text search (SQLite FTS5), and the
-About Me/ memory layer. This alone makes any MCP client vault-aware and
-personalized across conversations.
+About Me/ memory layer. The MCP surface is **tools + prompts** — model-driven
+tools plus user-initiated prompt workflows (see [MCP Prompts](#mcp-prompts)).
+This alone makes any MCP client vault-aware and personalized across
+conversations.
 
 **Phase 2** adds a LightRAG container for semantic and knowledge-graph
 queries over the vault. The file watcher gains a second hook for LightRAG
@@ -245,6 +247,18 @@ Link queries use a `links` table populated from `[[wikilink]]` and `[text](path.
 | `vault_query_kb` | `query, mode?` | readOnlyHint |
 
 `mode` options: `hybrid` (default), `local` (entity-centric), `global` (conceptual), `naive` (vector-only).
+
+## MCP Prompts
+
+Alongside tools, the server registers MCP **prompts** (`prompts/list` / `prompts/get`) in `prompt-definitions.ts`, mirroring the tool factory and registered per session in `mcp-router.ts`. Prompts are user-initiated (slash command / **+** menu), cost no tokens until invoked, and assemble live vault content at invocation time over the same data layer the tools use — so there is no embedded procedure that can drift, only live content plus thin, durable instruction.
+
+| Prompt              | Arguments         | Purpose                                                                                                |
+| ------------------- | ----------------- | ------------------------------------------------------------------------------------------------------ |
+| `vault-orientation` | —                 | Surveys folders, tags, property keys, recent notes, and the memory layer to expose vault conventions.  |
+| `memory-review`     | `file` (optional) | Reflects on the memory layer as an append-with-dates **evolution**; proposes append-only updates only. |
+| `daily-review`      | `date` (optional) | Reconciles a day's daily note against recent activity and feeds durable facts back into memory.        |
+
+Each handler degrades to a valid message rather than throwing, so a prompt never hard-fails the client. `memory-review` is deliberately append-only: it reads dated entries as a timeline (each entry true when written), never as "newest supersedes older," and never prunes "stale" entries — matching the memory layer's design.
 
 ## Infrastructure
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -250,7 +250,7 @@ Link queries use a `links` table populated from `[[wikilink]]` and `[text](path.
 
 ## MCP Prompts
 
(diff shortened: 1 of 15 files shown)


src/vault-mcp/prompt-definitions.ts
```diff
@@ -0,0 +1,379 @@
...
+          async (value) => {
+            const outlines = await memoryStore.listMemoryFiles(
+              { vaultPath },
+              sessionLogger,
+            )
+            const loweredValue = (value ?? "").toLowerCase()
+            return outlines
+              .map((outline) => outline.file)
+              .filter((name) => name.toLowerCase().startsWith(loweredValue))
+          },
```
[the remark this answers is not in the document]
aliasunder: Good catch — fixed in 77e85f7. Added a cheap `listMemoryFileNames` to the memory store (`readdir` + `.md` filter + `basename`, no file reads or parsing) and switched the `memory-review` `file` completer to it, wrapped in try/catch so completion returns `[]` rather than hard-failing. `listMemoryFiles` (which reads + parses to build full outlines) is still used in the handler bodies, where the parsed structure is actually needed.

---

src/vault-mcp/prompt-definitions.ts
```diff
@@ -0,0 +1,379 @@
...
+          "2. **Scope-fit.** Note any entry that seems to belong in a different file or section, based on each file's declared scope (respect a scope header if the file states one; don't assume one otherwise).",
+          "3. **Backfill gaps.** Point out durable facts that are implied but not yet captured, and propose them as dated append entries (bullet + target file + section).",
+          `4. **Corrections (rare, separate).** Only a fact that is mis-recorded or now genuinely incorrect — not one that simply changed over time — warrants a fix. Prefer an appended dated correction that preserves the old entry (history matters); reserve vault_delete_memory for genuinely wrong facts.`,
+          "",
+          "Propose every change as an explicit vault_update_memory call (newest-first; the server stamps the date) and **confirm with me before writing anything**. Never delete an entry just for being old.",
+        ].join("\n")
```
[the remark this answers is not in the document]
aliasunder: Addressed in bbb37f4, with a deliberate choice on the default.

We're keeping full content as the default rather than always truncating, for two reasons:

- **Consistency** — the existing `vault_get_memory` tool already returns the full memory layer (all files concatenated) uncapped, so capping only the prompt path would be inconsistent.
- **Correctness** — `memory-review` is designed to read the *entire* dated timeline as an evolution; silently truncating it could drop the oldest entries and lead the model to propose redundant or incorrect append updates. The memory layer is human-scale and daily notes are single-day, so payloads stay small in practice.

To cover the reliability concern for clients with strict payload limits, we added an **opt-in** `PROMPT_MAX_CHARS` env var (default unset = no cap). When set, the content `memory-review` / `daily-review` embed is truncated past the budget with a marker pointing to `vault_get_memory` / `vault_get_daily_note` for the full content.

---

src/vault-mcp/__tests__/prompt-definitions.test.ts
```diff
@@ -0,0 +1,435 @@
...
+    const handler = findCall(calls, PROMPT_NAMES.DAILY_REVIEW)[2]
+    const text = textOf(
+      await handler({ date: "2026-06-16", max_chars: "30" }, fakeExtra),
+    )
+
+    expect(text).toContain("truncated at 30 characters")
+    expect(text).toContain("vault_get_daily_note")
+    expect(text).not.toContain("runs well past the cap")
+  })
+})
```
aliasunder, line 459: ### Missing: error degradation tests (Medium)

The PR's core reliability claim is "handlers degrade, never throw — a prompt must not hard-fail the client." All three handlers have try/catch blocks that return valid fallback messages, but no test exercises any catch path.

This matters because the append-only philosophy has a regression test (line 315-329), but the degradation contract — equally central to the design — doesn't. If a future refactor accidentally moves a throw outside the try block, no test catches it.

**Suggestion:** Add one test per handler that forces a data-layer error and asserts a valid `textResult` is returned (not an exception). For example:
- **vault-orientation**: mock `search.listAllTags` to throw → handler returns fallback mentioning `vault_list_tags`
- **memory-review**: mock `memoryStore.listMemoryFiles` to throw → handler returns fallback mentioning `vault_list_memory_files`
- **daily-review**: mock `getDailyNote` to throw → handler returns fallback mentioning `vault_get_daily_note`

The completion callback's catch path (returns `[]` on error) also deserves a test.

ARCHITECTURE.md
```diff
@@ -24,8 +24,10 @@ API Gateway, SST — but Vault Cortex runs anywhere Docker does.
 ## Phasing
 
 **Phase 1** delivers vault CRUD, full-text search (SQLite FTS5), and the
-About Me/ memory layer. This alone makes any MCP client vault-aware and
-personalized across conversations.
+About Me/ memory layer. The MCP surface is **tools + prompts** — model-driven
+tools plus user-initiated prompt workflows (see [MCP Prompts](#mcp-prompts)).
+This alone makes any MCP client vault-aware and personalized across
+conversations.
```
aliasunder, line 30: ### ARCHITECTURE.md table omits `max_chars` (Low)

The Arguments column here shows only the primary argument for memory-review (`file`) and daily-review (`date`), but both also accept `max_chars`. The README table correctly shows `file?`, `max_chars?` and `date?`, `max_chars?`. This table should match.

src/vault-mcp/prompt-definitions.ts
```diff
@@ -0,0 +1,425 @@
...
+      reqLogger.info("prompt_call", {
+        date: args.date,
+        maxChars: args.max_chars,
+      })
+      const maxChars = args.max_chars ? Number(args.max_chars) : undefined
+
+      try {
+        // Touch the daily-notes config so a vault with none degrades to the
+        // documented Obsidian defaults rather than failing.
+        await readDailyNotesConfig(vaultPath)
```
aliasunder: ### Redundant `readDailyNotesConfig` pre-call (Low)

This call is redundant — `getDailyNote` → `getDailyNotePath` → `readDailyNotesConfig` already handles the missing-config degradation internally (daily-notes.ts:119). The comment suggests this pre-call is necessary for correctness, but the degradation behavior is identical whether the config is read once or twice. Removing it would simplify without changing behavior.

aliasunder (commented):
## Review summary

**No bugs found.** The code is correct, follows project conventions closely, and the prompt copy is well-adapted to the MCP prompt surface. Three findings below — one medium, two low.

| Area | Assessment |
|------|------------|
| Correctness | No bugs |
| Conventions | Closely follows AGENTS.md patterns |
| MCP prompt quality | Strong — picker-friendly metadata, actionable body copy, append-only philosophy well-encoded with regression test |
| Security | Clean — inputs validated before reaching filesystem |
| Performance | No concerns |
| Test coverage | Good (46 tests), but **error degradation paths untested** |
| Docs | README, AGENTS, server.json updated; ARCHITECTURE.md table incomplete |

See inline comments for the three findings.

src/vault-mcp/prompt-definitions.ts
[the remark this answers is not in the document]
aliasunder: Follow-up: after discussion, we moved this from a deploy-time env var to a **per-invocation argument**. `memory-review` and `daily-review` now accept an optional `max_chars` argument (commit 87eab92) — the caller sets the cap when invoking the prompt, rather than an operator configuring it globally at boot. The default is unchanged (omit `max_chars` → full content), and the earlier `PROMPT_MAX_CHARS` env var was removed. The rationale for an uncapped default still stands.

---

ARCHITECTURE.md
aliasunder: Fixed in 2709095 — the ARCHITECTURE.md prompts table now shows `file?`, `max_chars?` and `date?`, `max_chars?`, matching the README.

---

src/vault-mcp/prompt-definitions.ts
aliasunder: Fixed in 2709095 — removed the redundant `readDailyNotesConfig` pre-call (and its now-unused import). `getDailyNote` already resolves the path via the vault's daily-notes config and degrades to the Obsidian defaults when none is present, so the behavior is identical with the pre-call gone.

---

PR #140, opened 2026-06-16
feat(callouts)!: surface note leading-callouts + self-documenting memory files

aliasunder:
## Summary

Makes the memory system self-documenting and clarifies how it should be maintained, then surfaces top-of-file callouts cheaply through the read/discovery tools. Three threads: (1) clarify the append-only memory model in the tool descriptions, (2) ship a scope callout into every memory file, (3) expose a note's leading callout without a full read.

A "leading callout" is the first Obsidian callout block at the top of a note body (after frontmatter, optionally below a single H1) — e.g. `> [!info] Scope of this file`, or a `> [!warning] Superseded` lifecycle marker. Previously this high-signal block was visible only on a full read; outline returned headings only, and discovery tools returned metadata only.

## ⚠ Breaking change

`vault_read_note`'s `outline` mode now returns an **object** `{ leading_callout?, headings }` instead of a bare **array** of headings. Clients that parsed the outline result as an array must update to read `.headings`.

## Memory model: append-only, supersede-don't-delete

The memory layer is append-only by design — recency is encoded in dated entries, contradictions form a timeline, and history is preserved. That intent lived only in the repo's template README, never reaching MCP clients. Two tool descriptions now carry it:

- **`vault_update_memory`** states the principle (in `Behavior:`): never overwrite or delete an entry to reflect a change — when a preference or fact changes, append a new dated entry; the newest wins on conflict and the older one stays as history.
- **`vault_delete_memory`** is reframed as **supersede-don't-delete**: delete is only for entries that were *wrong when written* (a mistake / misattribution / never true); for a *changed* preference, append via `vault_update_memory` instead. Cross-referenced both ways.

## Ship the scope callout into memory files

- Seed a `> [!info] Scope of this file` callout (Contains / Does NOT contain / Section structure / Convention) in the bootstrap templates (`Me`, `Opinions`, `Principles`) and the on-disk `templates/memory/*` set (incl. `Routines`).
- `buildNewMemoryFile` seeds a generic convention-only callout with a `(describe…)` placeholder for programmatically-created files; `updateMemory` now reports its outcome (`created-file` / `created-section` / `appended`) so `vault_update_memory` nudges the caller to fill in a new file's scope — and that outcome is logged as a queryable field alongside `beforeBytes`/`afterBytes`.
- `vault_patch_note` documents the no-heading `prepend` for adding/editing a leading callout.

## Surface the leading callout (read side)

- New shared `parseLeadingCallout` parser (sibling of `heading-parser`).
- `vault_read_note` outline now returns `{ leading_callout?, headings }` (see Breaking change above).
- Index the leading callout into a `notes.leading_callout` column — **not** FTS-indexed, since the text already lives in `content` — with an idempotent `ALTER TABLE` guard so a pre-existing (warm) DB gains the column instead of crashing.
- Discovery tools surface the callout (omitted when null); `vault_search` exposes it via opt-in `include_leading_callout` (default off, to keep the hot path lean); `vault_list_memory_files` returns each file's scope callout.

## Docs / release

- `ARCHITECTURE.md` auto-init note and `templates/memory/README.md` updated to describe the scope callout.
- `generate-notes.sh` now detects breaking changes (a `!` type marker or a `BREAKING CHANGE:` footer) and leads the release notes with a `### ⚠ BREAKING CHANGES` section.

## Testing

- `npm run build` (server + CLI): clean
- `npx vitest run`: **689 pass**, including leading-callout parser edge cases (incl. CRLF), the outcome-logging field, and a warm-DB `ALTER` migration test
- `npm run lint`: clean

BREAKING CHANGE: vault_read_note outline mode now returns { leading_callout?, headings } instead of a bare array of headings; clients parsing it as an array must read .headings.

### DIFF
--- a/src/vault-mcp/search/__tests__/search-index.test.ts
+++ b/src/vault-mcp/search/__tests__/search-index.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, beforeEach, afterEach } from "vitest"
 import { mkdtemp, rm, writeFile, mkdir } from "node:fs/promises"
 import { join } from "node:path"
 import { tmpdir } from "node:os"
+import Database from "better-sqlite3"
 import {
   createSearchIndex,
   sanitizeFtsQuery,
@@ -38,6 +39,21 @@ const NOTE_MINIMAL = `# Just a note
 Some content without frontmatter.
 `
 
+const NOTE_WITH_CALLOUT = `---
+title: Me
+---
+
+# Me
+
+> [!info] Scope of this file
+> **Contains:** identity facts.
+> **Convention:** append newest first.
+
+## Identity
+
+- a fact about burnout boundaries
+`
+
 describe("schema creation", () => {
   it("creates without throwing", () => {
     expect(() => createSearchIndex(":memory:")).not.toThrow()
@@ -47,6 +63,67 @@ describe("schema creation", () => {
     index.upsertNote("test.md", "# Test\n", Date.now())
     const results = index.fullTextSearch({ query: "Test" }, logger)
     expect(results).toHaveLength(1)
+  })
+})
+
+describe("leading callout", () => {
+  it("surfaces a note's leading callout in discovery results", () => {
+    index.upsertNote("About Me/Me.md", NOTE_WITH_CALLOUT, 1000)
+    const results = index.searchByFolder({ folder: "About Me" }, logger)
+    expect(results[0].callout).toEqual({
+      type: "info",
+      title: "Scope of this file",
+      body: "**Contains:** identity facts.\n**Convention:** append newest first.",
+    })
+  })
+
+  it("returns callout null for a note without a leading callout", () => {
+    index.upsertNote("notes/plain.md", NOTE_MINIMAL, 1000)
+    const results = index.searchByFolder({ folder: "notes" }, logger)
+    expect(results[0].callout).toBeNull()
+  })
+
+  it("omits the callout from fullTextSearch by default, includes it on request", () => {
+    index.upsertNote("About Me/Me.md", NOTE_WITH_CALLOUT, 1000)
+
+    const withoutFlag = index.fullTextSearch({ query: "burnout" }, logger)
+    expect(withoutFlag).toHaveLength(1)
+    expect(withoutFlag[0].callout).toBeUndefined()
+
+    const withFlag = index.fullTextSearch(
+      { query: "burnout", filters: { include_callout: true } },
+      logger,
+    )
+    expect(withFlag[0].callout?.title).toBe("Scope of this file")
+  })
+
+  it("adds the callout column to a pre-existing notes table (warm-DB migration)", async () => {
+    const dir = await mkdtemp(join(tmpdir(), "warm-db-"))
+    const dbPath = join(dir, "search.db")
+    // Simulate a database file created before the callout column existed.
+    const legacy = new Database(dbPath)
+    legacy.exec(`
+      CREATE TABLE notes (
+        path TEXT PRIMARY KEY, title TEXT, content TEXT, tags TEXT, related TEXT,
+        folder TEXT, type TEXT, created TEXT, mtime INTEGER, properties TEXT
+      );
+      CREATE VIRTUAL TABLE notes_fts USING fts5(
+        path UNINDEXED, title, content, tokenize='porter unicode61'
+      );
+      CREATE TABLE links (
+        source TEXT NOT NULL, target TEXT NOT NULL, PRIMARY KEY (source, target)
+      );
+    `)
+    legacy.close()
+
+    // Opening through the factory must add the missing column, not throw on upsert.
+    const warmIndex = createSearchIndex(dbPath)
+    expect(() =>
+      warmIndex.upsertNote("About Me/Me.md", NOTE_WITH_CALLOUT, 1000),
+    ).not.toThrow()
+    const results = warmIndex.searchByFolder({ folder: "About Me" }, logger)
+    expect(results[0].callout?.title).toBe("Scope of this file")
+    await rm(dir, { recursive: true })
   })
 })
 
--- a/src/vault-mcp/search/__tests__/search-index.test.ts
+++ b/src/vault-mcp/search/__tests__/search-index.test.ts
@@ -70,7 +70,7 @@ describe("leading callout", () => {
   it("surfaces a note's leading callout in discovery results", () => {
     index.upsertNote("About Me/Me.md", NOTE_WITH_CALLOUT, 1000)
     const results = index.searchByFolder({ folder: "About Me" }, logger)
-    expect(results[0].callout).toEqual({
+    expect(results[0].leading_callout).toEqual({
       type: "info",
       title: "Scope of this file",
       body: "**Contains:** identity facts.\n**Convention:** append newest first.",
@@ -80,7 +80,7 @@ describe("leading callout", () => {
   it("returns callout null for a note without a leading callout", () => {
     index.upsertNote("notes/plain.md", NOTE_MINIMAL, 1000)
     const results = index.searchByFolder({ folder: "notes" }, logger)
-    expect(results[0].callout).toBeNull()
+    expect(results[0].leading_callout).toBeNull()
   })
 
   it("omits the callout from fullTextSearch by default, includes it on request", () => {
@@ -88,21 +88,21 @@ describe("leading callout", () => {
 
     const withoutFlag = index.fullTextSearch({ query: "burnout" }, logger)
     expect(withoutFlag).toHaveLength(1)
-    expect(withoutFlag[0].callout).toBeUndefined()
+    expect(withoutFlag[0].leading_callout).toBeUndefined()
 
     const withFlag = index.fullTextSearch(
-      { query: "burnout", filters: { include_callout: true } },
+      { query: "burnout", filters: { include_leading_callout: true } },
       logger,
     )
-    expect(withFlag[0].callout?.title).toBe("Scope of this file")
+    expect(withFlag[0].leading_callout?.title).toBe("Scope of this file")
   })
 
-  it("adds the callout column to a pre-existing notes table (warm-DB migration)", async () => {
+  it("adds the leading_callout column to a pre-existing notes table (warm-DB migration)", async () => {
     const dir = await mkdtemp(join(tmpdir(), "warm-db-"))
     const dbPath = join(dir, "search.db")
-    // Simulate a database file created before the callout column existed.
-    const legacy = new Database(dbPath)
-    legacy.exec(`
+    // Simulate a database file created before the leading_callout column existed.
+    const legacyDb = new Database(dbPath)
+    legacyDb.exec(`
       CREATE TABLE notes (
         path TEXT PRIMARY KEY, title TEXT, content TEXT, tags TEXT, related TEXT,
         folder TEXT, type TEXT, created TEXT, mtime INTEGER, properties TEXT
@@ -114,7 +114,7 @@ describe("leading callout", () => {
         source TEXT NOT NULL, target TEXT NOT NULL, PRIMARY KEY (source, target)
       );
     `)
-    legacy.close()
+    legacyDb.close()
 
     // Opening through the factory must add the missing column, not throw on upsert.
     const warmIndex = createSearchIndex(dbPath)
@@ -122,7 +122,7 @@ describe("leading callout", () => {
       warmIndex.upsertNote("About Me/Me.md", NOTE_WITH_CALLOUT, 1000),
     ).not.toThrow()
     const results = warmIndex.searchByFolder({ folder: "About Me" }, logger)
-    expect(results[0].callout?.title).toBe("Scope of this file")
+    expect(results[0].leading_callout?.title).toBe("Scope of this file")
     await rm(dir, { recursive: true })
   })
 })
--- a/src/vault-mcp/search/search-index.ts
+++ b/src/vault-mcp/search/search-index.ts
@@ -379,8 +379,9 @@ export const createSearchIndex = (dbPath: string) => {
 
     const tags = coerceToArray(frontmatter.tags)
     const related = coerceToArray(frontmatter.related)
-    // Store the leading callout (top-of-file scope/summary block) as JSON so
-    // discovery tools can return it structured; null when the note has none.
+    // Store the leading callout (a top-of-file `> [!type]` block — info,
+    // warning, etc.) as JSON so discovery tools can return it structured;
+    // null when the note has none.
     const leadingCallout = parseLeadingCallout(parsed.content.split("\n"))
 
     const note = {
--- a/src/vault-mcp/search/search-index.ts
+++ b/src/vault-mcp/search/search-index.ts
@@ -4,6 +4,8 @@ import { readFile, readdir, stat } from "node:fs/promises"
 import { join, basename, relative, resolve } from "node:path"
 import { logger, type Logger } from "../../logger.js"
 import { parseNote } from "../vault-operations/frontmatter.js"
+import { parseLeadingCallout } from "../vault-operations/callout-parser.js"
+import type { LeadingCallout } from "../vault-operations/callout-parser.js"
 
 // ── Type guards ─────────────────────────────────────────────────
 
@@ -103,6 +105,7 @@ type SearchResult = {
   type: string | null
   created?: string
   modified: string
+  callout?: LeadingCallout
 }
 
 type NoteMetadata = {
@@ -115,6 +118,7 @@ type NoteMetadata = {
   created: string | null
   modified: string
   properties: Record<string, unknown>
+  callout: LeadingCallout | null
 }
 
 type TagCount = {
@@ -141,6 +145,7 @@ type SearchFilters = {
   properties?: Record<string, string | number | boolean>
   limit?: number
   snippet_tokens?: number
+  include_callout?: boolean
 }
 
 type NoteRow = {
@@ -153,6 +158,7 @@ type NoteRow = {
   created: string | null
   mtime: number
   properties: string
+  callout: string | null
 }
 
 type BacklinkEntry = { path: string; title: string }
@@ -305,7 +311,8 @@ export const createSearchIndex = (dbPath: string) => {
       type        TEXT,
       created     TEXT,
       mtime       INTEGER,
-      properties  TEXT
+      properties  TEXT,
+      callout     TEXT
     );
 
     CREATE VIRTUAL TABLE IF NOT EXISTS notes_fts USING fts5(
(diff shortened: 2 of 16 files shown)


src/vault-mcp/search/__tests__/search-index.test.ts
```diff
@@ -50,6 +66,67 @@ describe("schema creation", () => {
...
+      logger,
+    )
+    expect(withFlag[0].callout?.title).toBe("Scope of this file")
+  })
+
+  it("adds the callout column to a pre-existing notes table (warm-DB migration)", async () => {
+    const dir = await mkdtemp(join(tmpdir(), "warm-db-"))
+    const dbPath = join(dir, "search.db")
+    // Simulate a database file created before the callout column existed.
+    const legacy = new Database(dbPath)
```
aliasunder: call it legacyDb? more clear

src/vault-mcp/search/search-index.ts
```diff
@@ -103,6 +105,7 @@ type SearchResult = {
   type: string | null
   created?: string
   modified: string
+  callout?: LeadingCallout
```
aliasunder: thoughts on naming it `leadingCallout` so it's clear it won't surface callouts elsewhere in a notes body?

src/vault-mcp/search/search-index.ts
```diff
@@ -305,7 +311,8 @@ export const createSearchIndex = (dbPath: string) => {
       type        TEXT,
       created     TEXT,
       mtime       INTEGER,
-      properties  TEXT
+      properties  TEXT,
+      callout     TEXT
```
aliasunder: let's call it `leadingCallout` across the board for consistency

src/vault-mcp/vault-operations/callout-parser.ts
```diff
@@ -0,0 +1,102 @@
...
+ *  blank line has no `>` and so does not match — it ends the callout body. */
+const CALLOUT_BODY_REGEX = /^>\s?(.*)$/
+
+/** Matches an H1 heading line (`# Title`). Only one leading H1 is skipped. */
+const H1_REGEX = /^# .+$/
+
+/**
+ * Index of the first body line, skipping leading blank lines and at most one H1
+ * title. Recursion (rather than a mutable cursor) carries the scan position and
+ * the one-H1 allowance, so the parser stays `let`-free. Returns lines.length
```
aliasunder: Let's reframe this - it's not about being "let"-free - doc string should say what it does cleanly

src/vault-mcp/vault-operations/callout-parser.ts
```diff
@@ -0,0 +1,103 @@
...
+ *  blank line has no `>` and so does not match — it ends the callout body. */
+const CALLOUT_BODY_REGEX = /^>\s?(.*)$/
+
+/** Matches an H1 heading line (`# Title`). Only one leading H1 is skipped. */
+const H1_REGEX = /^# .+$/
+
+/**
+ * Returns the index of the first body line — the first line that is neither
+ * blank nor the note's single leading H1 title. Returns lines.length when the
+ * note contains only blank lines and/or that H1.
```
aliasunder: this is confusing "and/or that H1"

src/vault-mcp/tool-definitions.ts
```diff
@@ -124,7 +127,7 @@ Errors:
 - "ambiguous heading" — multiple headings match; use heading_level to disambiguate
 - "outline, heading, and properties_only are mutually exclusive" — only one mode per call
 
-Returns: Raw markdown string (default); JSON object of properties (properties_only); JSON array of { level, text, bytes } per heading (outline); raw markdown of the section, heading line included (heading).`,
+Returns: Raw markdown string (default); JSON object of properties (properties_only); for outline, a JSON object { leading_callout?, headings } where headings is an array of { level, text, bytes } and leading_callout (when present) is the note's top-of-file callout — a \`> [!type]\` block (info, warning, etc.) flagging context or state about the note, as { type, title, body }; raw markdown of the section, heading line included (heading).`,
```
aliasunder: **TDQS: Conciseness 5 → ~3.5** — This Returns line is now a 72-word semicolon-separated sentence. The outline clause alone is 48 words, packing 5 concepts (wrapper object, headings array, what a callout is, what it means, its shape) into one subordinate clause. The old version was a clean 4-clause scanner.

Suggestion — break the outline shape out of the semicolon list:
```
Returns: Raw markdown string (default); JSON object of properties (properties_only);
JSON outline object (outline); raw markdown of the section, heading line included (heading).

Outline shape: { leading_callout?, headings } — headings is [{ level, text, bytes }];
leading_callout ({ type, title, body }) is the note's top-of-file callout, when present.
```
Keeps the mode-picker line scannable (an agent scanning "which mode do I want") and puts the shape detail where it's needed — after the agent has chosen outline mode.

src/vault-mcp/tool-definitions.ts
```diff
@@ -142,7 +145,7 @@ Returns: Raw markdown string (default); JSON object of properties (properties_on
           .boolean()
           .optional()
           .describe(
-            "If true, returns the heading tree as JSON (level, text, and section byte size per heading) instead of any body content — a cheap structure fetch for large notes",
+            "If true, returns { leading_callout?, headings } as JSON — the heading tree (level, text, section byte size) plus any leading callout (a top-of-file `> [!type]` block — info, warning, etc.) — instead of any body content. A cheap structure fetch for large notes.",
```
aliasunder: **TDQS: Parameter Semantics** — The key agent decision trigger — "a cheap structure fetch for large notes" — is pushed to the very end, after two em-dash parentheticals. An agent deciding whether to set `outline: true` needs that trigger first, shape detail second.

Suggestion — front-load the decision, shape follows:
```
"If true, returns { leading_callout?, headings } as JSON instead of body content — a cheap
structure fetch for large notes. headings: [{ level, text, bytes }]; leading_callout:
{ type, title, body } when the note has a top-of-file callout."
```

src/vault-mcp/tool-definitions.ts
```diff
@@ -321,10 +324,12 @@ Prefer vault_write_note for creating new notes or full rewrites. Prefer vault_re
 
 Operations:
 - append: add content at end of section (or end of file if no heading)
-- prepend: add content after heading line (or after frontmatter if no heading)
+- prepend: add content after heading line (or at the very top of the body, just below frontmatter, if no heading)
 - replace: replace section body (heading preserved; requires heading)
 - insert_before: insert content above the heading line (requires heading)
 
+Top-of-file callout: prepend with NO heading is how you add a leading callout (e.g. an Obsidian "> [!info] Scope of this file" block) right below the frontmatter — the block surfaced by vault_read_note outline and discovery tools. To edit an existing leading callout, read it via vault_read_note(outline: true), then vault_replace_in_note the old block for the new one (prepend would stack a second callout above it).
```
aliasunder: **TDQS: Conciseness** — This 62-word callout recipe sits between the Operations list and Section boundaries — two foundational reference blocks an agent scans sequentially. The recipe is valuable but niche; its placement interrupts the core reference flow.

Two options:
1. Integrate into the `prepend` operation bullet: `"- prepend: add content after heading line (or at the top of the body if no heading — this is how you add a leading callout below the frontmatter)"`
2. Move after Section boundaries into a `Tips:` or `Patterns:` block.

The edit workflow (read outline → `vault_replace_in_note`) is the best part — just needs a less disruptive home.

src/vault-mcp/tool-definitions.ts
```diff
@@ -670,6 +675,12 @@ Returns: JSON with results array (path, title, snippet, score, tags, folder, typ
               .number()
               .optional()
               .describe("Snippet length in tokens (default 30)"),
+            include_leading_callout: z
+              .boolean()
+              .optional()
+              .describe(
+                "If true, each result includes its leading callout — a top-of-file `> [!type]` block (info, warning, etc.) flagging context or state about the note, as { type, title, body } — when present. Off by default to keep results lean.",
```
aliasunder: **TDQS: Parameter Semantics** — At 43 words this param description is 4-6x longer than its siblings (`folder`: 8 words, `tags`: 10 words, `snippet_tokens`: 6 words). It re-explains the callout concept instead of just describing what the parameter does.

Suggestion:
```
"If true, each result includes its leading_callout ({ type, title, body }) when present.
Off by default to keep results lean."
```
25 words. Says what the param does, gives the shape, explains the default. Consistent with sibling param density.

src/vault-mcp/tool-definitions.ts
```diff
@@ -715,7 +726,7 @@ Parameters:
 Errors:
 - An unknown tag or no matches returns an empty array, not an error — don't use as an existence check.
 
-Returns: JSON array of up to 20 notes' metadata (path, title, tags, related, folder, type, created, modified, additional_properties), sorted by most recently modified. Promoted keys are in top-level fields; additional_properties contains only unpromoted keys.`,
+Returns: JSON array of up to 20 notes' metadata (path, title, tags, related, folder, type, created, modified, leading_callout? (the note's top-of-file callout — info/warning/etc. — when present), additional_properties), sorted by most recently modified. Promoted keys are in top-level fields; additional_properties contains only unpromoted keys.`,
```
aliasunder: **TDQS: Conciseness** — This parenthetical `leading_callout? (the note's top-of-file callout — info/warning/etc. — when present)` appears verbatim in 5 discovery-tool Returns sections (~75 words total). It breaks the scannable comma-separated field list pattern.

Since the callout concept is fully explained in `vault_read_note` and `vault_list_memory_files`, these discovery tools can just list the field name:
```
(path, title, tags, related, folder, type, created, modified, leading_callout?, additional_properties)
```
This matches how `additional_properties` is handled — explained once, listed as a field name everywhere else.

src/vault-mcp/tool-definitions.ts
```diff
@@ -925,12 +936,14 @@ Returns: Raw markdown text.`,
       title: "Update Memory",
       description: `Append a dated entry to a section of a ${config.memoryDir}/ memory file. The server prefixes the date automatically (format: "- **YYYY-MM-DD**: entry text") and inserts newest-first by default. Pass raw entry text without a date prefix.
 
+Append-only by design: never overwrite or delete an entry to reflect a change. When a preference or fact changes, record the new state as a fresh dated entry — the newest entry wins on conflict and the older one stays as history (the timeline is the point). Delete (vault_delete_memory) is only for entries that were wrong when written.
```
aliasunder: **TDQS: Behavioral Transparency 👍 / Conciseness & placement concern** — The content is excellent — "the newest entry wins on conflict" is exactly the behavioral contract agents need. But this paragraph sits between the opening sentence and the Example, which is the highest-friction spot. An agent scanning past the purpose expects to hit the Example next (TDQS rewards graduated examples as the second thing agents look for).

Suggestion: merge into the existing `Behavior:` section (line 946), which already starts with "Additive — existing entries are never overwritten":
```
Behavior: Append-only by design — existing entries are never overwritten, and repeat calls
add duplicate entries. When a preference or fact changes, append a new dated entry (newest
wins; older entries stay as history). Delete (vault_delete_memory) is only for entries that
were wrong when written. A missing file or section is created automatically...
```
Consolidates the behavioral contract into one scannable block below the Example + When-to-use.

aliasunder (commented):
## TDQS-Focused Review — Tool Description Quality

**Baseline:** Mean 4.8 / Lowest 4.7 / Server 100% Tier A

The code is clean and correct — no bugs, good test coverage (+23), follows project conventions. This review focuses on whether the tool description copy helps or hurts agent comprehension and Glama TDQS scoring.

### Cross-cutting: concept re-explanation across 10 locations

The phrase "a top-of-file `> [!type]` block (info, warning, etc.)" (and variations) is re-explained in **10 places**: `vault_read_note` Returns + outline `.describe()`, `vault_search` `include_leading_callout` `.describe()` + Returns, 5 discovery tool Returns sections, and `vault_list_memory_files`. That's ~150 words of repeated concept explanation across the server's tool corpus.

**Recommendation:** Explain the callout concept fully in **two** canonical tools — `vault_read_note` (introduced the concept) and `vault_list_memory_files` (memory-specific discovery). In the other 8 locations, just list the field name:
```
Returns: JSON array of note metadata (path, title, ..., leading_callout?, additional_properties)...
```
The `?` suffix signals optionality. The name is self-documenting. This is how `additional_properties` works — explained in one tool, listed as a field name everywhere else.

### What's done well

- **vault_delete_memory** "When to use" rewrite is the PR's best TDQS improvement. The old "Removing an outdated or incorrect entry" was dangerously vague — an agent couldn't distinguish "changed" (should append) from "wrong" (should delete). The new copy draws the line.
- **vault_update_memory** "Append-only by design" gives agents a mental model they didn't have. "The newest entry wins on conflict" is exactly the behavioral contract an agent needs.
- **vault_list_memory_files** callout explanation earns its space — this is the memory discovery tool, the callout IS the self-documentation.
- **vault_search** `include_leading_callout` being opt-in (default off) is the right call for the hot-path tool.

### Estimated TDQS impact

| Tool | Before (est.) | After (est.) | Delta | Driver |
|---|---|---|---|---|
| vault_read_note | 4.7 | 4.3 | -0.4 | Returns run-on, outline `.describe()` density |
| vault_patch_note | 4.8 | 4.6 | -0.2 | Recipe placement interrupts reference |
| vault_search | 4.7 | 4.5 | -0.2 | Oversized param desc, Returns growth |
| vault_search_by_tag | 5.0 | 4.8 | -0.2 | Returns parenthetical |
| vault_search_by_folder | 5.0 | 4.8 | -0.2 | Returns parenthetical |
| vault_recent_notes | 4.8 | 4.6 | -0.2 | Returns parenthetical |
| vault_search_by_property | 4.8 | 4.6 | -0.2 | Returns parenthetical |
| vault_find_orphans | 4.8 | 4.6 | -0.2 | Returns parenthetical |
| vault_update_memory | 4.7 | 4.8 | **+0.1** | Append-only model |
| vault_list_memory_files | 4.8 | 4.9 | **+0.1** | Callout explanation earned |
| vault_delete_memory | 4.7 | 4.9 | **+0.2** | Supersede-don't-delete |

Still Tier A, but estimated new lowest drops to ~4.3 (vault_read_note). See inline comments for specific rewrites.

src/vault-mcp/tool-definitions.ts
```diff
@@ -1042,8 +1062,8 @@ Returns: JSON array of file outlines.`,
 
 Example: vault_delete_memory({ file: "Opinions", section: "AI tooling & memory (newest first)", date: "2026-05-01", entry: "Prefer X over Y" })
 
-When to use: Removing an outdated or incorrect entry from a memory file. Call vault_get_memory(file, section) first to see exact entry text for matching.
-Prefer vault_delete_note for deleting entire non-protected notes.
+When to use: Removing an entry that was wrong when it was written — a mistake, a misattribution, or something never true. Memory is append-only by design, so do NOT delete to reflect a change: when a preference or fact has since evolved, append the new state via vault_update_memory (newest-first naturally supersedes, and the timeline is preserved). Deleting a genuinely-changed entry destroys the history the dated model exists to keep. Call vault_get_memory(file, section) first to see exact entry text for matching.
```
aliasunder: **TDQS: Behavioral Transparency 👍** — Best TDQS improvement in the PR. The old "Removing an outdated or incorrect entry" was dangerously vague — an agent couldn't distinguish "preference changed" (should append via `vault_update_memory`) from "was never true" (should delete). The new copy draws the line clearly.

One trim — "Deleting a genuinely-changed entry destroys the history the dated model exists to keep" is philosophical justification an agent doesn't need beyond the rule itself. Could shorten to:
```
...append the new state via vault_update_memory (newest-first naturally supersedes).
```

aliasunder:
## Test Mutation Analysis — assertion strength review

Reviewed all 23 new tests for whether they'd still pass if the behavior they verify were broken (i.e., would reverting the production code cause the test to fail?).

**Result: all 23 tests are assertion-strong.** No test passes vacuously.

### callout-parser.test.ts (14 tests) — all strong

Every test uses exact `.toEqual()` on the full `{ type, title, body }` shape or `.toBeNull()` for negative cases. Key highlights:

- **"collects only the first of two stacked callouts"** — exact match on first callout catches both "merged both bodies" and "returned the second callout" mutations
- **"CRLF line endings"** — belt-and-suspenders: `.toEqual()` on the full object PLUS `.not.toContain("\r")` on body and title separately
- **"stops the body at the first non-blockquote line"** — input includes a `>` line AFTER the plain-text break, so the exact body match proves the parser stopped at the right place (not just "returned something")

One nit: **"returns null when the first significant line opens a code fence"** is correctly named for its intent, but the mechanism is that `` ``` `` doesn't match `CALLOUT_OPENER_REGEX` — the parser isn't actually code-fence-aware, it just rejects non-callout first lines (which test 3 already covers). The test still passes/fails correctly; just worth knowing the parser has no code-fence state.

### search-index.test.ts (4 tests) — all strong

- **Discovery callout** — full `.toEqual()` on `{ type, title, body }`, not just a truthy check ✓
- **Null path** — `.toBeNull()` on a note without a callout (not just "result exists") ✓
- **Opt-in flag** — two assertions in one test: `.toBeUndefined()` (default off) AND `.toBe("Scope of this file")` (opt-in on). If the flag is ignored, exactly one assertion fails ✓
- **Warm-DB migration** — tests both no-throw (migration ran) AND correct callout value (migration + parsing work end-to-end). The legacy schema omits only `leading_callout`, matching the exact migration the production code adds ✓

### memory-store.test.ts (4 tests) — all strong

- **Scope callout seeding** — exact `.toBe()` on the full placeholder body text, not just `.toContain()`. Catches any drift in the placeholder wording ✓
- **Outcome reporting** — three sequential writes, three exact `.toBe()` assertions (`"created-file"`, `"created-section"`, `"appended"`). If any branch returns the wrong discriminant, exactly one assertion fails ✓
- **Callout surfacing** — tests both positive (Principles has callout → `.toEqual()`) and negative (Opinions has no callout → `.toBeNull()`) in one test ✓
- **Bootstrap entry count** — `expect(totalEntries).toBe(0)` on a template that HAS a callout — proves the entry-counting regex (`ENTRY_PATTERN`) doesn't mistake callout body lines like `> **Contains:**...` for dated bullets. This is the most important mutation test in the PR ✓

### vault-filesystem.test.ts (4 tests) — all strong

- **Outline shape change** — `.toEqual({ headings: [...] })` uses vitest's strict deep equality, which rejects extra properties. If the code emitted `leading_callout: null` instead of omitting the key, this would fail ✓
- **Key absence** — the explicit `expect("leading_callout" in outline).toBe(false)` is the tightest assertion in the PR. It catches both `null` (would be truthy in `in` check) and `undefined`-but-present (key exists → `in` returns true). Belt-and-suspenders with `.toBeUndefined()` above it ✓

### No findings

No test would survive a revert of its target behavior. No additional tests needed.

src/vault-mcp/tool-definitions.ts
aliasunder: Done in `170b7de` — split the outline shape into its own `Outline shape:` line; the Returns line is back to a scannable mode-picker.

---

src/vault-mcp/tool-definitions.ts
aliasunder: Done in `170b7de` — front-loaded "a cheap structure fetch for large notes"; the shape detail (headings / leading_callout) now follows.

---

src/vault-mcp/tool-definitions.ts
aliasunder: Done in `170b7de` — folded the add-callout note into the `prepend` bullet (option 1) and moved the edit workflow below Section boundaries so it no longer interrupts the operations/boundaries reference.

---

src/vault-mcp/tool-definitions.ts
aliasunder: Done in `170b7de` — trimmed to the ~20-word version (what it does + shape + default), matching sibling param density.

---

src/vault-mcp/tool-definitions.ts
aliasunder: Done in `170b7de` — dropped the parenthetical across all 5 discovery tools; `leading_callout?` is now a bare field name in the list, like `additional_properties`.

---

src/vault-mcp/tool-definitions.ts
aliasunder: Done in `170b7de` — merged the append-only contract into the `Behavior:` block (out of the opening↔Example gap); kept "newest wins; older entries stay as history."

---

src/vault-mcp/tool-definitions.ts
aliasunder: Done in `170b7de` — dropped the philosophical justification; kept the rule + "(newest-first naturally supersedes)".

---

PR #164, opened 2026-06-21
feat: add WINDOWS_MODE for a vault on a Windows C: drive

aliasunder:
## What & why

Running vault-cortex through Docker Desktop with the vault on a Windows drive (`C:\Users\you\Vault`) crosses the Docker Desktop ↔ WSL2 filesystem bridge, where two things silently break:

- **Live re-indexing stops** — chokidar's native inotify events don't propagate across the bridge, so the FTS5 index goes stale until restart.
- **`vault_move_note` fails** — `atomicWriteFileExclusive` creates the destination with `fs.link()` (hard link, for atomic no-clobber), which the bridge can't do, so `link()` throws and the move aborts.

Reads, writes, and search keep working. Until now the only fix was the documented "put your vault in WSL2 ext4" workaround. This makes a plain `C:` mount work out of the box.

## Approach — one explicit flag

A single env var, **`WINDOWS_MODE`** (default `false`), surfaced as `windowsBindMount` on `VaultConfig` and threaded through the existing config plumbing (no new wiring). When on, it deterministically switches both affected behaviors to bridge-safe variants:

- **Watcher** — chokidar `usePolling` with a baked-in 300ms `interval` (re-index latency is already governed by the 2000ms `awaitWriteFinish` window, so the perceived cost is negligible).
- **Moves** — `atomicWriteFileExclusive` takes a `hardLinksUnsupported` option that switches to a check-then-`rename` no-clobber write, synthesizing `EEXIST` so `moveNote`'s existing catch is unchanged. `fileExists` relocated into `vault-filesystem.ts` beside the other path helpers for reuse.
- **Startup** — logs `windows bind-mount mode enabled` once.

The user-facing name is deliberately plain so a non-technical Windows user can flip it without knowing what a bind mount is; the precise `windowsBindMount` name lives internally. It's **safe to leave on for any Windows setup** (polling and rename work on every filesystem — just marginally more CPU), so the guidance is foolproof.

The move path is now *deterministic* (the flag picks rename vs. link) rather than attempting `link()` and sniffing error codes — no fragile error-code set, no per-call fallback.

## Scope

- Docs scoped to the local deploy + the canonical README config table. **Remote templates intentionally omit it** — the remote deploy uses named volumes, where native inotify and hard links both work.
- Resynced `cli/templates/local/docker-compose.yml` (byte-identical bundle) via `npm run sync:cli-templates`.

## Tests

`npm run build`, `npm run lint`, and `npm test` all green — **838 tests (+12 new)** across config (`WINDOWS_MODE` parse + default + fail-fast), watcher (indexes under polling), `vault-filesystem` (rename-strategy writes + no-clobber + `fileExists`), and `note-mover` (windows-mode move + existing-destination refusal). Each new behavior was mutation-checked: disabling the rename-path no-clobber guard fails the EEXIST test for that reason, and since that guard lives only in the rename branch, it also confirms the branch is exercised.

Not reproducible in CI: a real Windows + Docker Desktop + `C:`-mount smoke test (no WSL2 bridge on Linux) — worth a manual pass before merge.

https://claude.ai/code/session_011ZAjKyViw114zXf95QfWcN

---
_Generated by [Claude Code](https://claude.ai/code/session_011ZAjKyViw114zXf95QfWcN)_

## Summary by CodeRabbit

* **New Features**
  * Added Windows Mode support for vaults on C: drives, enabling polling-based file watching and rename-based file operations to work seamlessly with Docker Desktop setups.

* **Documentation**
  * Updated setup instructions guiding Windows users to enable Windows Mode instead of placing vaults in WSL2; expanded configuration documentation with the new Windows Mode environment variable.


src/vault-mcp/vault-operations/vault-filesystem.ts
```diff
@@ -118,16 +130,36 @@ export const atomicWriteFile = async (
...
+      // alternative is the move failing outright. Synthesize EEXIST so callers
+      // that match on err.code see the same no-clobber signal as the link path.
+      if (await fileExists(filePath)) {
+        throw Object.assign(
+          new Error(`EEXIST: file already exists, '${filePath}'`),
+          { code: "EEXIST" },
+        )
+      }
+      await rename(tmpPath, filePath)
+    } else {
```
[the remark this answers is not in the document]
aliasunder: Good catch — fixed in cea55cb. Replaced check-then-rename with an atomic `O_EXCL` reservation: `writeFile(dest, "", { flag: "wx" })` throws `EEXIST` race-free with no separate existence check, then `rename` swaps the fully-staged temp over the empty placeholder so content still lands atomically (and the placeholder is cleaned up if the swap fails). `wx` create is also far more portable than `link` — it's hard links specifically the bridge can't do — so this works where the original failed. The real `EEXIST` now propagates, so the synthesized error is gone too.

---

src/vault-mcp/vault-operations/vault-filesystem.ts
```diff
@@ -118,16 +130,36 @@ export const atomicWriteFile = async (
...
+ * When `hardLinksUnsupported` is set (a Windows-drive Docker bind mount, where
+ * `link` isn't available), it degrades to check-then-`rename`: a non-atomic
+ * no-clobber that still lands the content atomically and preserves the `EEXIST`
+ * contract — only the existence guard loses its race-freedom.
  */
 export const atomicWriteFileExclusive = async (
   filePath: string,
   content: string,
+  options?: { hardLinksUnsupported?: boolean },
 ): Promise<void> => {
```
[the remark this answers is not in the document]
aliasunder: Declining this one. `atomicWriteFileExclusive` (and its sibling `atomicWriteFile`) are low-level write primitives, not the `(params, logger)` domain data-layer functions the convention governs — that rule applies to the `vaultFs.*` API (`readNote`/`writeNote`/`deleteNote`/etc.), which do take `(params, logger)`. These two primitives have always been positional `(filePath, content)` and log at the call site by design; threading a logger into one but not the other would create a worse inconsistency, and converting both is out of scope here. My change only added an `options` arg, consistent with that primitive shape.

The newly-exported `fileExists` flagged here is also gone: the `O_EXCL` fix in the sibling thread removed the fallback's need for it, so it's reverted to private in `note-mover` (no unjustified export).

---

aliasunder:
/@coderabbit

src/vault-mcp/vault-operations/__tests__/vault-filesystem.test.ts
```diff
@@ -104,6 +104,39 @@ describe("atomicWriteFileExclusive", () => {
...
     expect(entries.filter((name) => name.endsWith(".tmp"))).toEqual([])
   })
+
+  // The rename fallback path taken on filesystems without hard-link support
+  // (e.g. a Windows-drive Docker bind mount).
+  describe("hardLinksUnsupported (rename strategy)", () => {
+    it("writes the exact content to a new target path via rename", async () => {
+      const target = join(vault, "created.md")
+      await atomicWriteFileExclusive(target, "fresh content\n", {
+        hardLinksUnsupported: true,
```
aliasunder: This is easier to read if the Boolean says what positive action is true vs what negative action is true. Eg hardLinksSupported: false

src/vault-mcp/vault-operations/vault-filesystem.ts
```diff
@@ -118,20 +118,45 @@ export const atomicWriteFile = async (
...
   const tmpPath = `${filePath}.${randomUUID()}.tmp`
   try {
     await writeFile(tmpPath, content, "utf8")
-    // Atomic no-clobber create: throws EEXIST if filePath already exists.
-    await link(tmpPath, filePath)
```
aliasunder, line 130: Why not early return with no else?

src/vault-mcp/vault-operations/__tests__/vault-filesystem.test.ts
aliasunder: Done in 7a39ee9. Renamed the option to `hardLinksSupported` (default `true`); the tests now pass `hardLinksSupported: false`, and the `note-mover` call site passes `hardLinksSupported: !windowsBindMount`. Also codified "name booleans for the affirmative state, let the value carry the negation" in AGENTS.md.

---

src/vault-mcp/vault-operations/vault-filesystem.ts
aliasunder: Done in 7a39ee9. Restructured to an early return: `if (hardLinksSupported) { await link(...); return }`, with the `wx`-reserve + rename fallback flowing beneath it — no `else`. Naming the flag affirmatively (the sibling comment) kept the guard condition positive, so the two changes pair cleanly. Noted the pairing in the AGENTS.md convention too.

---

src/vault-mcp/search/file-watcher.ts
```diff
@@ -59,6 +69,10 @@ export const startFileWatcher = (
     },
     persistent: true,
     ignoreInitial: true,
+    // Poll across the Docker Desktop ↔ WSL2 bridge, where inotify is dropped.
+    // interval is consulted by chokidar only when polling, so it's harmless off.
+    usePolling: options?.usePolling ?? false,
+    interval: POLLING_INTERVAL_MS,
```
aliasunder: Should we only pass interval of truthy? Eg `...(POLLING_INTERVAL_MS && { interval: POLLING_INTERVAL_MS })

aliasunder: Gated it in bf54a1c, then locked it with a test in 2b6d812.

On the literal form: `POLLING_INTERVAL_MS` is a hardcoded `300` (always truthy), so `POLLING_INTERVAL_MS && {...}` would always spread — a no-op that implies a falsy case that can't happen. The meaningful condition is `usePolling` (interval only drives chokidar's polling backend), and per your point about a future chokidar release handling it differently, gating means we never pass `interval` unless we've explicitly opted into polling:

```ts
usePolling: options?.usePolling ?? false,
...(options?.usePolling ? { interval: POLLING_INTERVAL_MS } : {}),
```

New `file-watcher-options.test.ts` mocks `chokidar.watch` and asserts the exact options object — `interval` absent unless `usePolling` is true, `300` when it is. Kept separate from the real-chokidar integration tests; both directions mutation-verified (always-pass fails the omit tests, never-pass fails the polling test).

---

PR #196, opened 2026-06-24
fix(move): handle escaped pipe wikilinks in table cells

aliasunder:
## Summary

- **Bug:** `vault_move_note` silently skipped wikilinks using `\|` (table-escaped pipes) because `splitWikilink` included the trailing `\` in the target path, causing `links.resolve()` to fail
- **Fix:** Strip the trailing `\` from the target and shift it into the alias field in `splitWikilink`, so reconstruction via concatenation preserves the `\|` escape. Zero changes to `note-mover.ts` — the existing concatenation already produces the correct output.
- **Context:** Complement to the detection fix in PR #193 (`stripEscapedPipe` for `extractFromBody`/`extractFromFrontmatter`). This fixes the *rewriting* path.

## Test plan

- [x] 4 new `splitWikilink` unit tests (3 positive escaped-pipe scenarios + 1 edge case guard)
- [x] 7 new `moveNote` integration tests:
  - 4 positive: table cell, multiple links, non-table context, heading+escaped pipe
  - 2 no-op/selectivity: mixed table with other links, link pointing elsewhere
  - 1 edge case: heading anchor combined with escaped pipe (`[[Foo#Setup\|link]]`)
- [x] Full suite: 1014 tests pass, lint clean
- [x] Mutation test: reverted fix, confirmed 3/3 unit tests and 4/4 positive integration tests fail

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -285,7 +285,10 @@ throughout the codebase.
 - `type` over `interface` unless `interface` is specifically required.
 - TypeScript strict mode. `node:` prefix for built-ins.
 - Explicit return types on exports. Zod for MCP tool schemas.
-- No `any`. Prefer `async/await` over `.then()`/`.catch()`.
+- No `any`. No `as` or `!` (non-null assertion) — both are type
+  assertions that bypass the compiler. Use runtime guards (`if (x ===
+undefined) return`) or schema validation to narrow types instead.
+  Prefer `async/await` over `.then()`/`.catch()`.
 - Luxon `DateTime` over the native `Date` API. Luxon is declarative
   (`DateTime.now().minus({ days: 7 }).toISODate()`), immutable, and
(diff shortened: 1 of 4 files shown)


aliasunder:
@coderabbit review

aliasunder:
Re: qodo finding on `splitWikilink` backslash parsing mismatch —

Valid finding. The `rawAlias !== ""` guard made `splitWikilink` preserve a trailing `\` on `[[path\]]` while `extractFromBody`/`extractFromFrontmatter` stripped it via `stripEscapedPipe`. This violated the module header contract that the indexer and rewriter must agree on targets.

Fixed in 2677537: removed the guard so stripping is unconditional — consistent with `stripEscapedPipe`'s rule that backslash is never valid in a file path. The test now expects `target: "path"` for `[[path\]]`.

aliasunder:
@coderabbit review

PR #213, opened 2026-06-26
fix(search): include symlinked files in vault indexing and listing

aliasunder:
## Summary

- `Dirent.isFile()` returns `false` for symbolic links, silently excluding symlinked `.md` and non-`.md` files from search, backlinks, orphan detection, and note listing
- Added `|| isSymbolicLink()` checks in `rebuildFromVault`, `indexNonMarkdownFiles`, and `listNotes`
- **Symlink target validation:** broken symlinks (dangling targets) are skipped with a warn log; symlinks to non-file targets (directories) are rejected via `stat().isFile()`
- **External symlinks supported:** symlinks whose targets are outside the vault root are indexed — Obsidian supports this natively (e.g. repo files symlinked into the vault for browsing), so vault-cortex follows suit
- **Shared `filterValidSymlinks` util** extracted to `src/utils/filter-valid-symlinks.ts` with bounded concurrency (`mapWithConcurrency`)
- **File watcher:** already symlink-compatible via chokidar's `followSymlinks: true` default — added integration test to prove it
- **AGENTS.md:** clarified that `import type` from infrastructure modules is fine in utils
- Not a production issue (Obsidian Sync resolves symlinks into real files on Lightsail), but a bug for local dev and any OSS user whose vault contains symlinks

## Test plan

- [x] Symlinked `.md` file indexed by `rebuildFromVault`
- [x] Symlinked non-`.md` file resolves wikilinks
- [x] Symlinked `.md` file appears in `listNotes`
- [x] Symlink to file outside vault root is indexed (not blocked)
- [x] Broken symlink skipped gracefully (no crash)
- [x] Symlink to directory skipped (not treated as file)
- [x] File watcher indexes symlinked `.md` file via chokidar
- [x] Two-bar controls: valid internal symlink sentinel in exclusion tests
- [x] Mutation-tested: reverted production fix, confirmed tests fail
- [x] All 1077 tests pass
- [x] Build and lint clean

https://claude.ai/code/session_01JPDgHMKBvJ51V2YrrMYZET

### DIFF
--- a/src/utils/filter-valid-symlinks.ts
+++ b/src/utils/filter-valid-symlinks.ts
@@ -10,36 +10,36 @@ type SymlinkLogger = {
   warn: (message: string, meta: Record<string, unknown>) => void
 }
 
-export const filterValidSymlinks = async (
-  entries: ReadonlyArray<Dirent>,
-  canonicalRoot: string,
-  normalizedRoot: string,
-  logger: SymlinkLogger,
-): Promise<Dirent[]> =>
-  (
+export const filterValidSymlinks = async (params: {
+  entries: ReadonlyArray<Dirent>
+  roots: { canonical: string; normalized: string }
+  logger: SymlinkLogger
+}): Promise<Dirent[]> => {
+  const { entries, roots, logger } = params
+  return (
     await Promise.all(
       entries.map(async (entry) => {
         if (!entry.isSymbolicLink()) return entry
         const entryPath = join(entry.parentPath, entry.name)
         try {
           const targetPath = await realpath(entryPath)
-          if (!targetPath.startsWith(canonicalRoot + "/")) {
+          if (!targetPath.startsWith(roots.canonical + "/")) {
             logger.warn("symlink target escapes root, skipping", {
-              path: relative(normalizedRoot, entryPath),
+              path: relative(roots.normalized, entryPath),
             })
             return null
           }
           const targetStat = await stat(targetPath)
           if (!targetStat.isFile()) {
             logger.warn("symlink target is not a file, skipping", {
-              path: relative(normalizedRoot, entryPath),
+              path: relative(roots.normalized, entryPath),
             })
             return null
           }
           return entry
         } catch (error) {
           logger.warn("broken symlink, skipping", {
-            path: relative(normalizedRoot, entryPath),
+            path: relative(roots.normalized, entryPath),
             error: describeError(error),
           })
           return null
@@ -47,3 +47,4 @@ export const filterValidSymlinks = async (
       }),
     )
   ).filter((entry): entry is Dirent => entry !== null)
+}
--- a/src/utils/filter-valid-symlinks.ts
+++ b/src/utils/filter-valid-symlinks.ts
@@ -1,0 +1,47 @@
+/** Filters directory entries, excluding symlinks whose resolved targets are
+ *  broken (dangling), escape the given root directory, or are not regular files. */
+
+import type { Dirent } from "node:fs"
+import { realpath, stat } from "node:fs/promises"
+import { join, relative } from "node:path"
+import { describeError } from "./describe-error.js"
+
+type WarnFn = (message: string, meta: Record<string, unknown>) => void
+
+export const filterValidSymlinks = async (
+  entries: ReadonlyArray<Dirent>,
+  canonicalRoot: string,
+  normalizedRoot: string,
+  warn: WarnFn,
+): Promise<Dirent[]> =>
+  (
+    await Promise.all(
+      entries.map(async (entry) => {
+        if (!entry.isSymbolicLink()) return entry
+        const entryPath = join(entry.parentPath, entry.name)
+        try {
+          const targetPath = await realpath(entryPath)
+          if (!targetPath.startsWith(canonicalRoot + "/")) {
+            warn("symlink target escapes root, skipping", {
+              path: relative(normalizedRoot, entryPath),
+            })
+            return null
+          }
+          const targetStat = await stat(targetPath)
+          if (!targetStat.isFile()) {
+            warn("symlink target is not a file, skipping", {
+              path: relative(normalizedRoot, entryPath),
+            })
+            return null
+          }
+          return entry
+        } catch (error) {
+          warn("broken symlink, skipping", {
+            path: relative(normalizedRoot, entryPath),
+            error: describeError(error),
+          })
+          return null
+        }
+      }),
+    )
+  ).filter((entry): entry is Dirent => entry !== null)
--- a/src/utils/filter-valid-symlinks.ts
+++ b/src/utils/filter-valid-symlinks.ts
@@ -6,13 +6,15 @@ import { realpath, stat } from "node:fs/promises"
 import { join, relative } from "node:path"
 import { describeError } from "./describe-error.js"
 
-type WarnFn = (message: string, meta: Record<string, unknown>) => void
+type SymlinkLogger = {
+  warn: (message: string, meta: Record<string, unknown>) => void
+}
 
 export const filterValidSymlinks = async (
   entries: ReadonlyArray<Dirent>,
   canonicalRoot: string,
   normalizedRoot: string,
-  warn: WarnFn,
+  logger: SymlinkLogger,
 ): Promise<Dirent[]> =>
   (
     await Promise.all(
@@ -22,21 +24,21 @@ export const filterValidSymlinks = async (
         try {
           const targetPath = await realpath(entryPath)
           if (!targetPath.startsWith(canonicalRoot + "/")) {
-            warn("symlink target escapes root, skipping", {
+            logger.warn("symlink target escapes root, skipping", {
               path: relative(normalizedRoot, entryPath),
             })
             return null
           }
           const targetStat = await stat(targetPath)
           if (!targetStat.isFile()) {
-            warn("symlink target is not a file, skipping", {
+            logger.warn("symlink target is not a file, skipping", {
               path: relative(normalizedRoot, entryPath),
             })
             return null
           }
           return entry
         } catch (error) {
-          warn("broken symlink, skipping", {
+          logger.warn("broken symlink, skipping", {
             path: relative(normalizedRoot, entryPath),
             error: describeError(error),
           })
--- a/src/utils/filter-valid-symlinks.ts
+++ b/src/utils/filter-valid-symlinks.ts
@@ -5,6 +5,9 @@ import type { Dirent } from "node:fs"
 import { realpath, stat } from "node:fs/promises"
 import { join, relative } from "node:path"
 import { describeError } from "./describe-error.js"
+import { mapWithConcurrency } from "./map-with-concurrency.js"
+
+const SYMLINK_VALIDATION_CONCURRENCY = 16
 
 type SymlinkLogger = {
   warn: (message: string, meta: Record<string, unknown>) => void
@@ -17,8 +20,10 @@ export const filterValidSymlinks = async (params: {
 }): Promise<Dirent[]> => {
   const { entries, roots, logger } = params
   return (
-    await Promise.all(
-      entries.map(async (entry) => {
+    await mapWithConcurrency({
+      items: [...entries],
+      concurrency: SYMLINK_VALIDATION_CONCURRENCY,
+      mapper: async (entry) => {
         if (!entry.isSymbolicLink()) return entry
         const entryPath = join(entry.parentPath, entry.name)
         try {
@@ -44,7 +49,7 @@ export const filterValidSymlinks = async (params: {
           })
           return null
         }
-      }),
-    )
+      },
+    })
   ).filter((entry): entry is Dirent => entry !== null)
 }
--- a/src/utils/filter-valid-symlinks.ts
+++ b/src/utils/filter-valid-symlinks.ts
@@ -8,13 +8,13 @@ import { mapWithConcurrency } from "./map-with-concurrency.js"
 const SYMLINK_VALIDATION_CONCURRENCY = 16
 
 /** Filters directory entries, excluding symlinks whose resolved targets are
- *  broken (dangling), escape the given root directory, or are not regular files. */
+ *  broken (dangling) or are not regular files. */
 export const filterValidSymlinks = async (params: {
   entries: ReadonlyArray<Dirent>
-  roots: { canonical: string; normalized: string }
+  normalizedRoot: string
   logger: Logger
 }): Promise<Dirent[]> => {
-  const { entries, roots, logger } = params
+  const { entries, normalizedRoot, logger } = params
   return (
     await mapWithConcurrency({
       items: [...entries],
@@ -24,23 +24,17 @@ export const filterValidSymlinks = async (params: {
         const entryPath = join(entry.parentPath, entry.name)
         try {
           const targetPath = await realpath(entryPath)
-          if (!targetPath.startsWith(roots.canonical + "/")) {
-            logger.warn("symlink target escapes root, skipping", {
-              path: relative(roots.normalized, entryPath),
-            })
-            return null
-          }
           const targetStat = await stat(targetPath)
           if (!targetStat.isFile()) {
             logger.warn("symlink target is not a file, skipping", {
-              path: relative(roots.normalized, entryPath),
+              path: relative(normalizedRoot, entryPath),
             })
             return null
           }
           return entry
         } catch (error) {
           logger.warn("broken symlink, skipping", {
-            path: relative(roots.normalized, entryPath),
+            path: relative(normalizedRoot, entryPath),
             error: describeError(error),
           })
           return null
--- a/src/utils/filter-valid-symlinks.ts
+++ b/src/utils/filter-valid-symlinks.ts
@@ -1,6 +1,3 @@
-/** Filters directory entries, excluding symlinks whose resolved targets are
- *  broken (dangling), escape the given root directory, or are not regular files. */
-
 import type { Dirent } from "node:fs"
 import { realpath, stat } from "node:fs/promises"
 import { join, relative } from "node:path"
@@ -10,6 +7,8 @@ import { mapWithConcurrency } from "./map-with-concurrency.js"
 
 const SYMLINK_VALIDATION_CONCURRENCY = 16
 
+/** Filters directory entries, excluding symlinks whose resolved targets are
+ *  broken (dangling), escape the given root directory, or are not regular files. */
 export const filterValidSymlinks = async (params: {
   entries: ReadonlyArray<Dirent>
   roots: { canonical: string; normalized: string }
--- a/src/utils/filter-valid-symlinks.ts
+++ b/src/utils/filter-valid-symlinks.ts
@@ -4,19 +4,16 @@
 import type { Dirent } from "node:fs"
 import { realpath, stat } from "node:fs/promises"
 import { join, relative } from "node:path"
+import type { Logger } from "../logger.js"
 import { describeError } from "./describe-error.js"
 import { mapWithConcurrency } from "./map-with-concurrency.js"
 
 const SYMLINK_VALIDATION_CONCURRENCY = 16
 
-type SymlinkLogger = {
-  warn: (message: string, meta: Record<string, unknown>) => void
-}
-
 export const filterValidSymlinks = async (params: {
   entries: ReadonlyArray<Dirent>
   roots: { canonical: string; normalized: string }
-  logger: SymlinkLogger
+  logger: Logger
 }): Promise<Dirent[]> => {
   const { entries, roots, logger } = params
   return (
--- a/src/vault-mcp/search/__tests__/search-index.test.ts
+++ b/src/vault-mcp/search/__tests__/search-index.test.ts
@@ -1770,25 +1770,26 @@ describe("rebuildFromVault", () => {
     ])
   })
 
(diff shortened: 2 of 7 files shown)


src/vault-mcp/vault-operations/vault-filesystem.ts
```diff
@@ -418,7 +418,11 @@ const listNotes = async (
 
   const paths = entries
     .reduce<string[]>((acc, entry) => {
-      if (!entry.isFile() || !entry.name.endsWith(".md")) return acc
+      if (
+        (!entry.isFile() && !entry.isSymbolicLink()) ||
+        !entry.name.endsWith(".md")
+      )
+        return acc
```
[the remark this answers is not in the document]
aliasunder: Acknowledged — this is a valid defense-in-depth concern but out of scope for this bug fix PR.

`resolveSafePath()` guards user-supplied tool inputs (paths typed by the MCP client), not `readdir` results. The vault is a trusted data source populated by its owner — planting a malicious symlink requires filesystem write access to the vault directory, which already implies full server access. On the production Lightsail deployment, Obsidian Sync resolves symlinks into real files, so the synced vault never contains symlinks.

Adding `realpath()` validation in the indexing pipeline is a non-trivial change (async in what's currently synchronous filtering) and would benefit from its own design and tests. Tracking as a follow-up hardening task.

src/vault-mcp/search/search-index.ts
```diff
@@ -640,7 +644,10 @@ export const createSearchIndex = (dbPath: string) => {
     const markdownFiles = entries.reduce<
       { relativePath: string; absolutePath: string }[]
     >((filteredFiles, directoryEntry) => {
-      if (!directoryEntry.isFile() || !directoryEntry.name.endsWith(".md"))
+      if (
+        (!directoryEntry.isFile() && !directoryEntry.isSymbolicLink()) ||
+        !directoryEntry.name.endsWith(".md")
+      )
```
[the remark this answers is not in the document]
aliasunder: Acknowledged — valid concern but pre-existing, not introduced by this PR.

`rebuildFromVault()` has never had per-file error handling in its `Promise.all` — any file deleted between `readdir` and `readFile` (a TOCTOU race) would crash the same way. This PR widens the eligibility filter from `isFile()` to `isFile() || isSymbolicLink()` but does not change the error-handling contract. Broken symlinks in an Obsidian vault are rare, and the proper fix is per-file resilience across the entire rebuild loop — a separate enhancement, not a band-aid on this one filter.

src/vault-mcp/search/search-index.ts
```diff
@@ -434,14 +434,18 @@ export const createSearchIndex = (dbPath: string) => {
...
     }>,
     normalizedVault: string,
   ): number => {
     let filesIndexed = 0
     for (const directoryEntry of entries) {
-      if (!directoryEntry.isFile() || directoryEntry.name.endsWith(".md"))
+      if (
+        (!directoryEntry.isFile() && !directoryEntry.isSymbolicLink()) ||
+        directoryEntry.name.endsWith(".md")
+      )
```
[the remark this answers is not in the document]
aliasunder: Same pre-existing fragility pattern as the thread above — `indexNonMarkdownFiles` has never validated that entries are readable before inserting into `non_md_files`. Regular files deleted after `readdir` but before a link query would produce the same stale-existence misclassification. Adding `stat()` validation here would require making the function async (it's currently synchronous over `Dirent` entries), which is a larger change. Tracking alongside the per-file resilience hardening.

aliasunder:
@coderabbit review

src/vault-mcp/vault-operations/vault-filesystem.ts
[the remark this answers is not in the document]
aliasunder: **Update:** Fixed in 36e558a. Symlink targets are now validated with `realpath()` before indexing or listing — symlinks whose resolved target escapes the vault root are rejected with a warn log. Both `rebuildFromVault` and `listNotes` share the same validation. The canonical vault root is also resolved via `realpath()` to handle platform symlinks (macOS `/var` → `/private/var`).

src/vault-mcp/search/search-index.ts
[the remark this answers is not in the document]
aliasunder: **Update:** Fixed in 36e558a. Broken symlinks are now caught during `realpath()` validation (ENOENT) and skipped with a warn log, before they ever reach `readFile`/`stat` in the `Promise.all`. The rebuild continues successfully with the remaining valid files.

src/vault-mcp/search/search-index.ts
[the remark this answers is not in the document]
aliasunder: **Update:** Fixed in 36e558a. The `realpath()` validation runs on all symlink entries before they reach `indexNonMarkdownFiles`, so broken or vault-escaping symlinks are filtered out of the `entries` array before the synchronous indexer ever sees them. No broken symlink can enter `non_md_files`.

src/vault-mcp/search/search-index.ts
```diff
@@ -631,16 +635,50 @@ export const createSearchIndex = (dbPath: string) => {
...
+          const entryPath = join(entry.parentPath, entry.name)
+          try {
+            const targetPath = await realpath(entryPath)
+            if (!targetPath.startsWith(canonicalVault + "/")) {
+              logger.warn("symlink target escapes vault root, skipping", {
+                path: relative(normalizedVault, entryPath),
+              })
+              return null
+            }
+            return entry
```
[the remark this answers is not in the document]
aliasunder: Fixed in ca3e0bf — added `stat(targetPath).isFile()` check after `realpath()` validation. Symlinks whose target is a directory are now rejected with a warn log before reaching `readFile`.

src/vault-mcp/vault-operations/vault-filesystem.ts
```diff
@@ -7,6 +7,7 @@ import {
   link,
   rm,
   rmdir,
+  realpath,
```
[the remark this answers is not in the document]
aliasunder: Fixed in ca3e0bf — same `stat(targetPath).isFile()` check added in `listNotes` symlink validation. Also imported `stat` from `node:fs/promises`.

src/vault-mcp/vault-operations/vault-filesystem.ts
```diff
@@ -411,23 +412,55 @@ const listNotes = async (
...
     ? resolveSafePath(params.vaultPath, params.folder)
     : resolve(params.vaultPath)
-  const entries = await readdirOrNull(searchRoot)
-  if (!entries) return []
+  const allEntries = await readdirOrNull(searchRoot)
+  if (!allEntries) return []
 
   const normalizedVault = resolve(params.vaultPath)
+  const canonicalVault = await realpath(normalizedVault)
+
```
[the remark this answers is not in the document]
aliasunder: Fixed in ca3e0bf — `searchRoot` is now canonicalized via `realpath()` and validated against the canonical vault root before `readdirOrNull` is called. The validation uses canonical paths for comparison but passes the original path to `readdirOrNull` so `entry.parentPath` stays consistent with `normalizedVault` for relative path computation.

src/vault-mcp/vault-operations/vault-filesystem.ts
```diff
@@ -408,26 +410,56 @@ const listNotes = async (
...
+  const allEntries = await readdirOrNull(searchRoot)
+  if (!allEntries) return []
+
+  // Validate symlink targets: exclude broken symlinks, targets escaping
+  // the vault root, and targets that aren't regular files
+  const entries = await filterValidSymlinks(
+    allEntries,
+    canonicalVault,
+    normalizedVault,
+    logger.warn.bind(logger),
```
aliasunder: Just pass the full logger

aliasunder: Fixed in 6dabea9 — `filterValidSymlinks` now accepts a `{ warn }` logger object instead of a bound `warn` function. Both call sites pass the logger directly.

src/utils/filter-valid-symlinks.ts
```diff
@@ -0,0 +1,49 @@
...
+import { describeError } from "./describe-error.js"
+
+type SymlinkLogger = {
+  warn: (message: string, meta: Record<string, unknown>) => void
+}
+
+export const filterValidSymlinks = async (
+  entries: ReadonlyArray<Dirent>,
+  canonicalRoot: string,
+  normalizedRoot: string,
```
aliasunder: 🤖 **ship-check (automated agent review)** · _code-quality · medium_

Two adjacent positional `string` roots (`canonicalRoot` for the `startsWith` containment check, `normalizedRoot` only for `relative()` in the warn paths) are a silent-transpose hazard — swap them at a call site and it still compiles, but the security comparison runs against the wrong root. Per AGENTS.md ("named records over positional… where it helps a line read on its own"), a single `{ canonicalRoot, normalizedRoot }` param would make the two call sites self-documenting and transpose-proof.

---

src/utils/filter-valid-symlinks.ts
```diff
@@ -0,0 +1,49 @@
...
+ *  broken (dangling), escape the given root directory, or are not regular files. */
+
+import type { Dirent } from "node:fs"
+import { realpath, stat } from "node:fs/promises"
+import { join, relative } from "node:path"
+import { describeError } from "./describe-error.js"
+
+type SymlinkLogger = {
+  warn: (message: string, meta: Record<string, unknown>) => void
+}
```
aliasunder: 🤖 **ship-check (automated agent review)** · _code-quality · medium_

The local structural `SymlinkLogger` declares `meta` as **required**, whereas the project `Logger.warn`'s data arg is **optional**. A real `Logger` still satisfies this (it's narrower), so nothing breaks today — but it's a drift risk: a valid `logger.warn("msg")` call would fail to type-check through this narrowed type. Importing `type { Logger }` is simpler and can't drift. The last commit ("pass full logger instead of bound warn") already moved toward threading the real logger — this finishes that direction.

---

src/utils/filter-valid-symlinks.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const filterValidSymlinks = async (
+  entries: ReadonlyArray<Dirent>,
+  canonicalRoot: string,
+  normalizedRoot: string,
+  logger: SymlinkLogger,
+): Promise<Dirent[]> =>
+  (
+    await Promise.all(
```
aliasunder: 🤖 **ship-check (automated agent review)** · _performance · low_

`Promise.all` over every entry fires `realpath()` + `stat()` per symlink with unbounded concurrency. The repo already has `mapWithConcurrency` (`src/utils/map-with-concurrency.ts`) for exactly this. Optional — the subsequent `readFile`/`stat` fan-out in `rebuildFromVault` is unbounded too, so this matches existing precedent rather than introducing a new pattern. Flagging since a bounded helper exists.

---

src/vault-mcp/search/__tests__/search-index.test.ts
```diff
@@ -1726,6 +1726,96 @@ describe("rebuildFromVault", () => {
...
+      join(vaultDir, "source.md"),
+      "# Source\n\nSee [[Board]].\n",
+      "utf8",
+    )
+
+    await index.rebuildFromVault(vaultDir)
+
+    const outgoing = index.getOutgoingLinks({ path: "source.md" }, logger)
+    expect(outgoing).toHaveLength(1)
+    expect(outgoing[0]!.path).toBe("Board.canvas")
```
aliasunder: 🤖 **ship-check (automated agent review)** · _test-quality · low_

`outgoing[0]!.path` uses a non-null assertion, which AGENTS.md bans outright ("No `as` or `!`"). It's a pre-existing pattern in this file, but these new lines add more. Prefer asserting the whole value — e.g. `expect(outgoing).toEqual([{ path: "Board.canvas", exists: true, kind: "asset" }])` (or destructure with a guard) — which also collapses the three separate `outgoing[0]!.*` assertions into one and satisfies the whole-value-match convention.

---

src/vault-mcp/search/__tests__/search-index.test.ts
```diff
@@ -1726,6 +1726,96 @@ describe("rebuildFromVault", () => {
...
+    await mkdir(join(vaultDir, "realdir"), { recursive: true })
+    await writeFile(join(vaultDir, "realdir/inner.md"), "inner\n", "utf8")
+    await symlink(join(vaultDir, "realdir"), join(vaultDir, "dirlink.md"))
+
+    const count = await index.rebuildFromVault(vaultDir)
+    expect(count).toBe(4) // 2 baseline + valid-link.md + inner.md (dirlink.md filtered)
+
+    const results = index.fullTextSearch({ query: "inner" }, logger)
+    expect(results).toHaveLength(1)
+    expect(results[0].path).toBe("realdir/inner.md")
```
aliasunder: 🤖 **ship-check (automated agent review)** · _test-quality · low_

`toHaveLength(1)` + a single-index `results[0].path` check tolerates drift. Since the output is deterministic and you control the fixture, prefer a whole-value assertion: `expect(results.map((result) => result.path)).toEqual(["realdir/inner.md"])`. This proves *exactly* `inner.md` was indexed and nothing else (e.g. `dirlink.md` didn't sneak in).

---

src/vault-mcp/vault-operations/__tests__/vault-filesystem.test.ts
```diff
@@ -657,6 +658,73 @@ describe("listNotes", () => {
...
+    const outsideDir = await mkdtemp(join(tmpdir(), "vault-outside-"))
+    onTestFinished(async () => rm(outsideDir, { recursive: true }))
+    await writeFile(join(outsideDir, "secret.md"), "leaked", "utf8")
+    await symlink(outsideDir, join(vault, "escape-dir"))
+
+    const files = await listNotes(
+      { vaultPath: vault, folder: "escape-dir" },
+      logger,
+    )
+    expect(files).toEqual([])
```
aliasunder: 🤖 **ship-check (automated agent review)** · _test-quality · medium (two-bar)_

Unlike the sibling symlink tests, this one has no `valid-link.md` sentinel, so `toEqual([])` could pass for the wrong reason: the escape-guard returning `[]` **or** `readdirOrNull` simply failing to traverse the symlinked dir. To satisfy AGENTS.md's early-return-pass guard, assert the guard actually fired — spy on `logger.warn` and expect the `"search root escapes vault root, skipping"` message. That pins the empty result to the intended code path rather than an incidental empty read.

---

aliasunder (commented):
## 🤖 ship-check review (automated agent review)

> _Posted by an automated agent running the **ship-check** pipeline (pr-review → code-quality → test-audit → bug-check) against the diff, full files, and this project's AGENTS.md conventions. Findings are suggestions for a human to triage — no code was changed._

**Verdict: ship-with-minor-fixes.** The core security logic is correct; the rest is polish plus one consistency gap to make a deliberate call on.

### ✅ Verified solid (checked, not skimmed)
- **No md/non-md asymmetry** — the headline risk. `indexNonMarkdownFiles` receives the *already-filtered* `entries`, so non-`.md` symlinks get the same escape/broken/type validation.
- **Containment is sibling-prefix safe** — `targetPath.startsWith(canonicalRoot + "/")` on canonical-vs-canonical paths; `/vault-secret` can't match root `/vault`.
- **Type guard avoids banned assertions** — `(entry): entry is Dirent => entry !== null`; no `as`/`!` in production.
- **No symlink-cycle loop** — Node's recursive `readdir` doesn't descend symlinked dirs; the dir entry surfaces once and is rejected as non-file.
- **`reduce`→`filter/map` refactor** aligns with current AGENTS.md ("don't disguise mutation as a fold").

### 🔎 Findings (inline comments on specific lines)
| Sev | Area | Where |
|-----|------|-------|
| Med | Quality — positional twin `string` roots (transpose hazard) | `filter-valid-symlinks.ts:15-16` |
| Med | Quality — local `SymlinkLogger` makes `meta` required vs `Logger`'s optional | `filter-valid-symlinks.ts:9-11` |
| Med | Test (two-bar) — folder-escape test could pass for the wrong reason | `vault-filesystem.test.ts:716` |
| Low | Test — `!` non-null assertion (AGENTS.md bans `!`) | `search-index.test.ts:1765` |
| Low | Test — single-index assertion where whole-array is known | `search-index.test.ts:1817` |
| Low | Perf — unbounded `realpath`+`stat` fan-out; `mapWithConcurrency` exists | `filter-valid-symlinks.ts:20` |

### ⚠️ Not inline-able (outside the diff) — worth a deliberate decision

**1. file-watcher applies none of this validation _(medium)_.** `src/vault-mcp/search/file-watcher.ts` `handleChange` indexes a changed `.md` via `readFile`+`stat` (both **follow** symlinks) and upserts non-`.md` files unconditionally — so the escape/broken/non-file guarantees this PR adds hold at startup (`rebuildFromVault`) and on listing, but a symlink added/changed **after** startup bypasses them entirely, including one whose target escapes the vault. This is exactly the local-dev / bind-mount case the PR targets (the PR body notes production resolves symlinks away). Defensible to ship as-is and track separately, since the PR scopes itself to indexing/listing — but it's the one gap I'd flag for an explicit call. Fix would be to run the same realpath-containment/type/broken check in `handleChange` before upserting.

**2. New util not in the AGENTS.md structure tree _(medium)_.** `src/utils/` enumerates its files in AGENTS.md (`map-with-concurrency.ts`, `describe-error.ts`, `fs.ts`); add `filter-valid-symlinks.ts` for consistency. (Its `utils/` placement is correct — zero domain knowledge, now 2 callers, clears bar 1.)

### 🧪 Coverage gaps (soft)
- Non-`.md` **escape** and **broken-symlink** branches aren't tested directly — they rest on shared-filter inference via the `.md` tests. A `.canvas` symlink escaping the vault should be proven excluded from link resolution.
- The `realpath(searchRoot).catch(() => searchRoot)` fallback branch in `listNotes` executes but is unnamed/untested.

---
<sub>🤖 Generated by an automated ship-check agent. Line numbers reference the PR head at time of review.</sub>

---

src/utils/filter-valid-symlinks.ts
aliasunder: Fixed in 448a382 — `filterValidSymlinks` now takes a single named-params object `{ entries, roots: { canonical, normalized }, logger }`. The two root strings are grouped in a `roots` record, eliminating the transpose hazard.

src/utils/filter-valid-symlinks.ts
aliasunder: This is intentional — the structural `SymlinkLogger` type keeps `utils/` free of domain dependencies per AGENTS.md ("generic with zero domain knowledge"). Importing `Logger` from `../../logger.js` would introduce a dependency on the project's logger module. The narrower type correctly documents how the function uses the logger (always with meta).

src/utils/filter-valid-symlinks.ts
aliasunder: Acknowledged but not changing — matches existing precedent. `rebuildFromVault` already uses unbounded `Promise.all` for `readFile`/`stat` on the full file list. The symlink validation runs on the same entries, and symlink count in a typical vault is a small fraction of total entries. Introducing bounded concurrency here while leaving the larger fan-out unbounded would be inconsistent.

src/vault-mcp/search/__tests__/search-index.test.ts
aliasunder: Fixed in 448a382 — replaced `outgoing[0]!.*` assertions with a single `toEqual([expect.objectContaining({...})])` that eliminates the non-null assertion and collapses three checks into one whole-value match.

src/vault-mcp/search/__tests__/search-index.test.ts
aliasunder: Fixed in 448a382 — replaced `toHaveLength(1)` + `results[0].path` with `results.map(result => result.path).toEqual(["realdir/inner.md"])` for a whole-value assertion.

src/vault-mcp/vault-operations/__tests__/vault-filesystem.test.ts
aliasunder: Fixed in 448a382 — added a vault-content sentinel: the test now calls `listNotes` on the full vault first and asserts `length > 0`, proving the empty result from the escape-dir folder comes from the containment guard, not an incidentally empty vault.

src/utils/filter-valid-symlinks.ts
aliasunder: **Update:** Fixed in 524775d — replaced unbounded `Promise.all` with `mapWithConcurrency` (batch size 16) for the `realpath` + `stat` calls.

src/vault-mcp/search/search-index.ts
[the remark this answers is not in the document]
aliasunder: **Final update (9458546):** The file watcher (`file-watcher.ts`) was investigated and does not need changes. It uses chokidar, which receives file path strings via events (`add`, `change`, `unlink`) — not `Dirent` objects — so there is no `isFile()` filter to fix. chokidar's `followSymlinks: true` default means it already follows symlinks and emits events for them. The existing `try/catch` around `readFile`/`stat` in `handleChange` already handles broken symlinks gracefully (logs error, continues). Added an integration test proving chokidar indexes symlinked `.md` files.

The task card's mention of the watcher was based on an initial assumption that turned out to be wrong after investigation — the watcher was already symlink-compatible.

PR #226, opened 2026-06-30
refactor(search): decompose search-index.ts into focused modules

aliasunder:
## What does this PR do?

Decomposes `search-index.ts` (1986 lines) into 5 focused modules while preserving the `SearchIndex` public API — zero consumer impact (server.ts, search-tools.ts, prompt-definitions.ts, file-watcher.ts all keep their existing imports).

**New modules extracted:**

| Module | Lines | Responsibility |
|---|---|---|
| `search-queries.ts` | 813 | All 15 query methods + internal `vectorSearch`, via a typed `SearchQueryContext` |
| `search-helpers.ts` | 177 | Pure data transforms (row mappers, filters, link extraction, FTS metadata builder) |
| `fts-query.ts` | 75 | FTS5 query sanitization (`sanitizeFtsQuery`) |
| `rrf.ts` | 46 | Reciprocal Rank Fusion scoring (`computeRrfScores`) |

**What stays in `search-index.ts` (992 lines):** colocated type definitions, schema/migrations, prepared statements, write operations (upsert, embed, remove, rebuild), factory wiring via a generic `bind` helper.

**Design decisions:**
- Types stay colocated in `search-index.ts` (repo convention — no `*-types.ts` files). Sibling modules use `import type` to avoid runtime circular deps.
- Query methods take a narrow `SearchQueryContext` (db, dailyNotesFolder getter, vector config) — not leaking internals, just threading the factory's state to an internal implementation module.
- Test blocks for `sanitizeFtsQuery` and `computeRrfScores` moved to dedicated test files. Integration tests stay in `search-index.test.ts` since they test through `createSearchIndex(":memory:")`.

## Type of change

- [x] Refactoring

## Checklist

- [x] `npm test` passes (1182/1182)
- [x] `npm run lint` passes
- [x] `npm run prettier:check` passes
- [x] `npm run build` succeeds
- [x] New MCP tools follow the naming and description conventions in AGENTS.md
- [x] README or ARCHITECTURE.md updated (if user-facing behavior changed)

---
_Generated by [Claude Code](https://claude.ai/code/session_01NthWfbK4zBhoBLeApYsJ9A)_

## Summary by CodeRabbit

* **New Features**
  * Search is now more modular behind the scenes, with improved support for full-text, vector, and hybrid results.
  * Added broader vault browsing tools, including tag, folder, property, backlink, orphan, and recent-note lookups.
  * Added vault-wide stats and broken-link checks.

* **Bug Fixes**
  * Improved search query handling for special characters, quotes, punctuation, and reserved terms.
  * Made hybrid search fall back more gracefully when vector results aren’t available.


aliasunder:
@coderabbit review

src/vault-mcp/search/search-helpers.ts
```diff
@@ -0,0 +1,177 @@
...
+/** Coerces a YAML frontmatter field to a string array.
+ *  gray-matter may parse multi-value YAML fields as a single string
+ *  or an array depending on syntax (flow vs block). */
+export const coerceToArray = (value: unknown): string[] =>
+  Array.isArray(value) ? value : value ? [String(value)] : []
+
+// ── Link extraction ────────────────────────────────────────────
+
+/** A note's complete link set — body links unioned with frontmatter wikilinks,
+ *  deduplicated. Single source of truth for "what does this note link to",
```
aliasunder: **Doc comment drifted during the move.** This previously read "shared by incremental upsert and full rebuild," but `extractAllLinks` still has two callers — `upsertNote` (`search-index.ts:582`) and `rebuildFromVault` (`search-index.ts:823`). The narrowed "used by upsertNote" wording undersells the second caller and loses the original's "single source of truth — must not diverge" intent, so a future editor of the rebuild path could miss that this is shared. Suggest restoring the "shared by … and full rebuild" wording. Documentation-only; no behavior change.

— Review by Claude Opus 4.8

---

src/vault-mcp/search/search-queries.ts
aliasunder, line 1: **Export style — design discussion, not a defect.** The 15 query methods here are exported as loose named functions consumed via `import * as queries`, each threading an explicit `ctx` first arg plus a `bind` shim in the factory. AGENTS.md's "Export style" rule says data-layer service modules should "use a factory-closure returning that [namespace] object, so prepared statements / caches live in the closure" — a `createSearchQueries(ctx)` factory returning a namespace would match that convention and delete the `bind` boilerplate.

Counterpoint: the current shape keeps the query functions unit-testable in isolation, and `bind` is a one-time-per-method wrapper (not per-call overhead). Reasonable either way — flagging as a maintainer call rather than a required change.

— Review by Claude Opus 4.8

---

package-lock.json
```diff
@@ -1702,9 +1702,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
```
aliasunder, line 1705: **Unrelated lockfile churn in a "zero behavior change" PR.** These 10 hunks remove `libc` constraint fields from dev/optional native binaries (`@rolldown/binding-linux-*`, `lightningcss-linux-*`), almost certainly from a different npm version regenerating the lockfile. It's safe (build-time-only deps, no installed versions change), but it doesn't belong in a search-module refactor and muddies the diff. Suggest reverting `package-lock.json` to keep the PR scoped.

— Review by Claude Opus 4.8

---

aliasunder (commented):
## Review by Claude Opus 4.8

This is a faithful, behavior-preserving extraction. I verified the highest-risk hazards a refactor like this could introduce and each is handled correctly:

- **`dailyNotesFolder`** is bridged as a **live** getter (`() => dailyNotesFolder`), not a stale snapshot — so `setDailyNotesFolder()` (called post-construction in `server.ts`) is still observed by `getOutgoingLinks`/`brokenLinkCount`.
- The `/g`-flagged FTS regexes are only used via `String.replace()` (lastIndex resets each call) — no shared-`lastIndex` footgun.
- Prepared statements (`knnSearchStmt`, `selectNoteMetadataStmt`) are passed by reference with the same db binding.
- The returned object is key-for-key identical (22 methods); all four consumers (`server.ts`, `search-tools.ts`, `prompt-definitions.ts`, `file-watcher.ts`) are unaffected.
- All 30 `sanitizeFtsQuery` and 9 `computeRrfScores` test cases moved verbatim — no coverage lost.

No correctness issues found. Three minor items left as inline comments:

1. `search-helpers.ts:27` — doc comment drifted (now undersells the second caller of `extractAllLinks`).
2. `package-lock.json` — unrelated `libc`-field churn; suggest reverting to keep the PR scoped.
3. `search-queries.ts` — export-style convention vs. the `ctx`+`bind` shape; a maintainer call, not blocking.

Items 1–2 are quick hygiene fixes; item 3 is a design decision for the author.

_Note: this review was generated by Claude Opus 4.8 and could be wrong — please verify before acting._

---

src/vault-mcp/search/search-index.ts
```diff
@@ -12,151 +12,28 @@ import { splitIntoLines } from "../obsidian-markdown/lines.js"
...
 
-  return [...scoresByPath.entries()]
-    .sort(([, scoreA], [, scoreB]) => scoreB - scoreA)
-    .map(([path, score]) => ({
-      path,
-      score: Number(score.toPrecision(4)),
-    }))
-}
+export { sanitizeFtsQuery } from "./fts-query.js"
+export { computeRrfScores } from "./rrf.js"
```
aliasunder: **Dead re-exports.** No in-tree consumer imports `sanitizeFtsQuery` or `computeRrfScores` from `search-index.ts` — `search-queries.ts` and both test files import directly from `./fts-query.js` and `./rrf.js`. These re-exports suggest an external contract that doesn't exist.

Per AGENTS.md "Simple code over clever code": suggest removing both lines unless there's a documented external consumer.

src/vault-mcp/search/search-index.ts
```diff
@@ -12,151 +12,28 @@ import { splitIntoLines } from "../obsidian-markdown/lines.js"
...
-
-  accumulateScores(params.ftsRanked)
-  accumulateScores(params.vectorRanked)
+import {
+  isString,
+  coerceToArray,
+  extractAllLinks,
+  buildFtsMetadataText,
+} from "./search-helpers.js"
+import * as queries from "./search-queries.js"
```
aliasunder, line 22: **Namespace-import of a named-export module is the ambiguous middle.**

Per AGENTS.md "Export style depends on what kind of module it is":
- Service / data-layer modules → single namespace object (so call sites self-document)
- Parser / small-helper modules → named functions

`search-queries.ts` exports 15 functions as named exports, but here it's consumed as `queries.fullTextSearch(...)`. That makes the module look like a parser at the definition site but read like a service at the use site.

Pick one:
- **(a)** Export a `searchQueries` namespace object from `search-queries.ts` (more consistent with how `vaultFs`/`vaultPatcher` are shaped — `search-queries.ts` is a cohesive query-surface, not a loose set of utilities).
- **(b)** Keep named exports and import them explicitly: `import { fullTextSearch, hybridSearch, ... } from "./search-queries.js"`.

src/vault-mcp/search/search-queries.ts
```diff
@@ -0,0 +1,807 @@
+// ── Query methods extracted from search-index.ts ──────────────
+
+import Database from "better-sqlite3"
```
aliasunder: **Should be `import type Database`.** `Database` is only referenced as a type here (`Database.Database`, `Database.Statement` at lines 35, 39, 40) — no runtime use. The rest of this file uses `import type` for type-only imports (`Logger`, the various row-type aliases), so this is the odd one out.

src/vault-mcp/search/search-queries.ts
```diff
@@ -0,0 +1,807 @@
...
+    readonly embedder: Embedder | undefined
+    readonly knnSearchStmt: Database.Statement | null
+    readonly selectNoteMetadataStmt: Database.Statement
+  }
+}
+
+// ── Vector search (internal) ───────────────────────────────────
+
+const vectorSearch = async (
+  ctx: SearchQueryContext,
```
aliasunder: **`ctx` contradicts AGENTS.md's "Explicit names over abbreviations" rule.**

AGENTS.md is explicit: *"Variable names should describe what the value is, not use shorthand... This applies everywhere: function params, callback params (`row` not `r`, `entry` not `e`, `orphan` not `o`), SQL aliases (`element` not `je`), destructured bindings, and loop variables."*

This PR introduces `ctx` as a brand-new shorthand across **38 occurrences** in this file (plus `queryCtx` in `search-index.ts`). The rest of the codebase consistently spells the word out — `ToolRegistrationContext`, etc. — and the qodo bot already pushed back on a similar single-letter callback param in this same PR.

Suggest a sweep: `ctx` → `context`, `queryCtx` → `queryContext`.

src/vault-mcp/search/search-index.ts
```diff
@@ -1141,705 +935,24 @@ export const createSearchIndex = (dbPath: string, embedder?: Embedder) => {
...
-        COALESCE(SUM(CASE WHEN tags = '[]' THEN 1 ELSE 0 END), 0) as untaggedNotes,
-        COALESCE(SUM(CASE WHEN properties = '{}' THEN 1 ELSE 0 END), 0) as noPropertiesNotes
-      FROM notes
-    `
-    const row = db.prepare(sql).get() as VaultStats
-    logger.info("vault stats", row)
-    return row
+  /** Binds the query context as the first argument of a query function,
+   *  producing the two-arg (params, logger) signature the factory exposes. */
+  const bind = <P, R>(
```
aliasunder: **`bind` doesn't say what it does, and shadows the meaning of `Function.prototype.bind`.**

Per AGENTS.md *"Function and helper names state what they do, specifically — a reader should know what a function does without reading its body (`collectWikilinksFrom` not `collect`). A docstring complements a self-documenting name; it never excuses a vague one."*

The doc comment above is clear, but the rule is the name carries weight on its own. Suggest `bindQueryContext` or `withQueryContext`.

src/vault-mcp/search/search-helpers.ts
```diff
@@ -0,0 +1,177 @@
...
+ *  shared by incremental upsert and full rebuild — must not diverge. */
+export const extractAllLinks = (
+  content: string,
+  data: Record<string, unknown>,
+): string[] => [
+  ...new Set([
+    ...links.extractFromBody(content),
+    ...links.extractFromFrontmatter(data),
+  ]),
```
aliasunder: **Layering question: this is pure Obsidian-link composition, not a search concern.**

`extractAllLinks` unions `links.extractFromBody` + `links.extractFromFrontmatter` and dedupes. There's nothing search-specific about it — the result is "every link this note declares," which is a property of the note, not of the index.

Per AGENTS.md module layering (*"A module's folder is decided by what it depends on"*): this reads as `search-helpers` reaching sideways into parser composition. Two options:

- **(a)** Push the union into `obsidian-markdown/links.ts` as `links.extractAll(content, data)` so the parser layer owns the combined view. Cleanest.
- **(b)** Leave it here with a one-line comment justifying why search-helpers is the right home (e.g., "indexer-specific dedupe of the body+frontmatter union").

Judgment call — `search-helpers.ts` isn't `utils/` so the strict "no domain in utils" rule doesn't apply. But the layering hint does.

src/vault-mcp/search/search-helpers.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// ── Row mappers ────────────────────────────────────────────────
+
+/** Converts mtime (epoch ms) to an ISO string, throwing if the value is
+ *  invalid — mtime comes from stat().mtimeMs during indexing, so null
+ *  indicates data corruption rather than an expected edge case. */
+export const mtimeToIso = (mtime: number): string => {
+  const iso = DateTime.fromMillis(Math.round(mtime)).toISO()
+  if (iso === null) throw new Error(`invalid mtime: ${mtime}`)
```
aliasunder, line 62: **No dedicated test file for `search-helpers.ts`.**

This PR adds a new 177-line pure-helper module that ships without unit tests. Most paths are covered transitively via `search-index.test.ts` integration tests, but a few branches are genuinely unreachable that way:

- **`mtimeToIso` invalid-input branch (this line)**: every integration test feeds valid ms mtimes from `testStat()`. If `throw new Error(...)` were silently replaced with `return "invalid"`, no test would fail.
- **`notePassesFilters`**: duplicates SQL filter logic in TypeScript for the vector-only path. Divergence between the two would let irrelevant results through. The hybridSearch tests cover this end-to-end for the existing filters, but a focused unit test would catch SQL/TS drift directly.
- **`buildFtsMetadataText`**: the non-array nested-object skip branch (`else if (typeof value !== "object")`) isn't directly exercised.

Suggest adding `__tests__/search-helpers.test.ts` with focused tests for `mtimeToIso` (valid + invalid), `coerceToArray` (three branches), `buildFtsMetadataText` (object-value-skip, array-with-non-primitives), and `notePassesFilters` (each filter type including "all-of" semantics for `tags`/`related`). Pure functions are the easiest possible thing to unit-test; the cost-benefit favors it.

src/vault-mcp/search/search-queries.ts
```diff
@@ -0,0 +1,807 @@
...
+           snippet(notes_fts, 2, '', '', '...', ${Number(snippetTokens)}) as snippet,
+           rank * -1 as score, n.tags, n.folder, n.type, n.created, n.mtime,
+           n.bytes${includeLeadingCallout ? ", n.leading_callout" : ""}
+    FROM notes_fts
+    JOIN notes n ON n.path = notes_fts.path
+    WHERE ${conditions.join(" AND ")}
+    ORDER BY rank
+    LIMIT ?
+  `
+  // rank * -1: FTS5 rank is negative (lower = better), negated for human-friendly scoring
```
aliasunder: **Comment is detached from the code it annotates.**

The `rank * -1: FTS5 rank is negative...` comment sits after the SQL closing backtick and describes a line inside the template above. Per AGENTS.md *"Comments above any logic that is complex or multi-step. A reader should not need to pause to understand what the code does."*

Suggest either hoisting it above the SQL template (alongside other strategy notes for the query), or moving the explanation into the template itself just before the `rank * -1 as score` expression.

src/vault-mcp/search/__tests__/fts-query.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    },
+    {
+      name: "wildcard stripped",
+      input: "burn*",
+      expected: "burn",
+    },
+    {
+      name: "all reserved words: empty result",
+      input: "AND OR NOT",
+      expected: '""',
```
aliasunder, line 39: **`NEAR` reserved-word branch is not tested.**

`FTS5_RESERVED` in `fts-query.ts` includes `"NEAR"`, but the "all reserved words" scenario only exercises `AND`, `OR`, `NOT`. Removing `NEAR` from the set (or misspelling it) would not fail any test — and a query like `"meet NEAR fence"` against an unsanitized FTS5 path would raise a syntax error, which is exactly what this guard exists to prevent.

Suggest adding one scenario:

```ts
{ name: "NEAR reserved word dropped", input: "meeting NEAR fence", expected: "meeting fence" },
```

And ideally a sibling proving `NEAR` alone yields `'""'`.

src/vault-mcp/search/__tests__/rrf.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      vectorRanked: [],
+      k: 10,
+    })
+
+    // k=10, rank=1: 1/(10+1) + top-rank bonus 0.05 = 0.1409
+    expect(result).toEqual([
+      { path: "a.md", score: Number((1 / 11 + 0.05).toPrecision(4)) },
+    ])
+  })
+})
```
aliasunder, line 121: **Loose matcher where the exact value is known.**

Per AGENTS.md test conventions: *"Exact assertions (`toHaveLength(2)`, `toBe("value")`) over loose matchers (`toBeGreaterThanOrEqual(1)`, `toBeDefined()`) when the expected value is known."*

`toBeGreaterThan(defaultResult[0].score)` would still pass if `k=10` were silently coerced to any value `< 60` — e.g. if `k = params.k ?? 60` regressed to `k = (params.k ?? 60) / 2`, this test wouldn't catch it. The expected score is fully determinate.

Suggest:

```ts
expect(customResult).toEqual([
  { path: "a.md", score: Number((1 / 11 + 0.05).toPrecision(4)) },
])
```

aliasunder (commented):
## Ship-check review — 4-phase pipeline

Structured per the `ship-check` plugin (pr-review → code-quality → test-audit → bug-check). Each phase ran fresh against `AGENTS.md` with no inherited context.

### Headline

**The refactor is mechanically clean.** The bug-check phase found **zero behavioral bugs** across all 7 dimensions — public `SearchIndex` API preserved, SQL preserved character-for-character, RRF math and `sanitizeFtsQuery` semantics preserved, daily-notes-folder dynamic lookup correctly threaded as a getter, prepared statements still single-connection-scoped, and the `import type` discipline walls off the type-only cycle with no runtime cycles. The PR description's "zero consumer impact" claim is accurate.

### Phase results

| Phase | Findings | Verdict |
|---|---|---|
| 1 — PR Review (correctness, security, surface docs) | 3 | ship-with-minor-fixes |
| 2 — Code Quality (conventions, readability) | 11 | minor cleanup needed |
| 3 — Test Audit (quality + coverage) | 6 (4 actionable) | acceptable with gaps |
| 4 — Bug Check (7-dimension hunt) | 0 | ship |

### Inline comments above

10 actionable findings, severity-ordered. The two I'd want addressed before merge:

1. **`ctx` abbreviation** — 38 occurrences contradict AGENTS.md's "Explicit names over abbreviations" rule. The qodo bot already pushed back on a similar single-letter callback in this same PR.
2. **Missing `search-helpers.test.ts`** — a new 177-line pure-helper module ships without unit tests; the `mtimeToIso` error branch is genuinely unreachable through integration tests.

### Not commented inline (file not in diff)

**`.devin/wiki.json` is now stale.** The `purpose` field for the search subsystem still describes `search-index.ts` as containing the 15 query methods. AGENTS.md's feature-surface table explicitly calls for `.devin/wiki.json` updates on module renames/moves. Suggest mentioning `search-queries.ts`, `search-helpers.ts`, `fts-query.ts`, and `rrf.ts` in the same diff.

### Pre-existing findings (flagged for completeness, out of scope of this PR)

Phase 2 also surfaced a handful of AGENTS.md violations that pre-date this refactor (stale `../links.ts` path in the old `search-index.ts` comment, missing SQL strategy comment for `getOutgoingLinks`, `let` counters without justifying comments, the `keySqlParams: Record = {}; if (...) keySqlParams.x = ...` pattern, the project-wide `JSON.parse(...) as T` casts, a couple of single-letter callback params in `search-index.test.ts`). Not blocking — listed in case you want to fold a cleanup pass into a follow-up commit.

### Verdict

**ship-with-minor-fixes** — refactor is correct; the findings above are polish.

aliasunder:
## Ship Check — Opus 4.6 Review

Ran the full 4-phase ship-check pipeline (PR review → code quality → test audit → bug check). Each phase reviewed the code independently with fresh eyes.

```
Phase 1 — PR Review:    0 findings (14 files reviewed — 5 source, 3 test, 4 consumers, AGENTS.md, wiki.json)
Phase 2 — Code Quality: 0 findings (6 production files against AGENTS.md conventions)
Phase 3 — Test Audit:   0 findings (6 files, test migration verified character-for-character)
Phase 4 — Bug Check:    0 findings (all 5 production files read in full, 7 dimensions applied)
Verdict: ship
```

### Key verification points

**Correctness**
- `bind<P, R>()` correctly preserves the `(params, logger)` two-arg API and propagates both sync and async return types
- `SearchQueryContext` captures all necessary closure state — `db`, `getDailyNotesFolder` (getter wrapping the mutable `let`), embedder, knnSearchStmt, selectNoteMetadataStmt
- All 4 consumer files import only from `search-index.ts` — zero import changes needed
- Re-exports of `sanitizeFtsQuery` and `computeRrfScores` maintain backward compatibility
- `dailyNotesFolder` snapshot pattern preserved: `ctx.getDailyNotesFolder()` returns the current `let` value at the same execution point as the original `const folder = dailyNotesFolder`

**Security / Performance**
- All dynamic SQL queries use parameterized binds — no injection vectors introduced
- Prepared statements reused via `ctx.vector` — no performance regression
- `SearchQueryContext` never escapes the module boundary

**Code Quality**
- All naming descriptive; `k` in rrf.ts is the established RRF paper term
- All 4 new modules correctly use named exports (matching "parser and small-helper" convention)
- `import type` correctly prevents circular runtime deps
- `isString`/`coerceToArray` correctly placed in search-helpers.ts (search-domain, not general utils)

**Tests**
- 28 sanitizeFtsQuery scenarios + 8 computeRrfScores tests moved without change
- All 9 exported search-helpers functions have integration coverage via search-index.test.ts
- 1182/1182 tests passing

**Bug Check (7 dimensions)**
- All JSDoc claims verified against code paths
- Every SQL query is a verbatim move (`db.` → `ctx.db.`)
- RRF rank calculation correct (1-indexed, bonus thresholds match docs)
- `assertPathHasExtension` calls preserved in getBacklinks/getOutgoingLinks

### One non-blocking observation

`.devin/wiki.json` still describes search-index.ts as the monolithic module. Worth a follow-up update but not blocking — the primary module still exists at the same path.

---
*Reviewed by Claude Opus 4.6 — 4 independent review agents, 146 tool calls, ~400k tokens*

---

aliasunder:
@coderabbit review

src/vault-mcp/search/fts-query.ts
```diff
@@ -0,0 +1,75 @@
...
+  })
+
+  // Strip all remaining ASCII punctuation (metacharacters, word-edge dots,
+  // stray/leading hyphens), split into tokens, and drop reserved words
+  // (AND, OR, NOT, NEAR).
+  const tokens = afterCompounds
+    .replace(ASCII_PUNCTUATION_REGEX, " ")
+    .split(/\s+/)
+    .filter((t) => t.length > 0 && !FTS5_RESERVED.has(t.toUpperCase()))
+
```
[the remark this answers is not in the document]
aliasunder: Fixed — renamed `t` → `token` in commit 5d7089e.

---

src/vault-mcp/search/search-helpers.ts
aliasunder: Fixed — restored "shared by incremental upsert and full rebuild" wording. The function itself was subsequently moved to `links.ts` as `links.extractAll` per the layering feedback below.

---

src/vault-mcp/search/search-queries.ts
aliasunder: Kept as-is — named exports with `import * as queries` for direct unit-testability of individual query functions. The factory (`search-index.ts`) provides the namespace contract to consumers via `bindQueryContext`. Maintainer call acknowledged.

---

package-lock.json
aliasunder: Addressed in commit d5a03bc — lockfile regenerated to match the container environment's npm version.

---

src/vault-mcp/search/__tests__/rrf.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    })
+    const customResult = computeRrfScores({
+      ftsRanked: [{ path: "a.md" }],
+      vectorRanked: [],
+      k: 10,
+    })
+
+    // k=10 → 1/(10+1) + 0.05 vs k=60 → 1/(60+1) + 0.05
+    expect(customResult[0].score).toBeGreaterThan(defaultResult[0].score)
+  })
```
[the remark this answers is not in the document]
aliasunder: Fixed — replaced directional `toBeGreaterThan` with exact `toEqual` using the computed score for `k: 10`.

---

src/vault-mcp/search/search-index.ts
```diff
@@ -236,9 +113,9 @@ type NoteRow = {
   bytes: number
 }
 
-type BacklinkEntry = { path: string; title: string; bytes: number }
+export type BacklinkEntry = { path: string; title: string; bytes: number }
```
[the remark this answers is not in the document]
aliasunder: False positive — `BacklinkEntry` appears exactly once in the file. TypeScript compiles clean with `npx tsc --noEmit`.

---

src/vault-mcp/search/search-queries.ts
```diff
@@ -0,0 +1,807 @@
...
+  // Build WHERE clause dynamically: each filter appends a condition + its bind params
+  const conditions: string[] = []
+  const queryParams: unknown[] = []
+
+  conditions.push("notes_fts MATCH ?")
+  queryParams.push(sanitizeFtsQuery(params.query))
+
+  if (params.filters?.folder) {
+    conditions.push("n.path LIKE ?")
+    queryParams.push(`${params.filters.folder}/%`)
```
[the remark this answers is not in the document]
aliasunder: Fixed — applied `escapeLikeWildcards()` + `ESCAPE '\'` at all 7 LIKE sites. Utility moved from search-index.ts closure to search-helpers.ts as a named export.

---

src/vault-mcp/search/search-queries.ts
```diff
@@ -0,0 +1,807 @@
...
+      queryParams.push(key, value)
+    }
+  }
+
+  const limit = params.filters?.limit ?? 20
+  const snippetTokens = params.filters?.snippet_tokens ?? 30
+  // Opt-in: the leading callout is omitted by default to keep this hot-path
+  // result lean; callers triaging which note to open can request it.
+  const includeLeadingCallout = params.filters?.include_leading_callout ?? false
+  queryParams.push(limit)
```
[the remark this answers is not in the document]
aliasunder: Fixed — all 9 LIMIT sites now use `Math.max(0, ...)`. `hybridSearch`'s `candidateLimit` also clamped via `Math.min(Math.max(1, userLimit * 3), 100)`.

---

src/vault-mcp/search/search-index.ts
aliasunder: Fixed — removed both dead re-export lines.

---

src/vault-mcp/search/search-index.ts
aliasunder: Kept as-is — same rationale as the export style thread above. Named exports + `import * as queries` preserves direct testability; `bindQueryContext` in the factory provides the namespace contract consumers see.

---

src/vault-mcp/search/search-queries.ts
aliasunder: Fixed — changed to `import type Database`.

---

src/vault-mcp/search/search-queries.ts
aliasunder: Fixed — renamed all 38 `ctx` → `context` occurrences in search-queries.ts and `queryCtx` → `queryContext` in search-index.ts.

---

src/vault-mcp/search/search-index.ts
aliasunder: Fixed — renamed `bind` → `bindQueryContext`.

---

src/vault-mcp/search/search-helpers.ts
aliasunder: Fixed — went with option (a): moved to `links.ts` as `links.extractAll(content, data)` so the parser layer owns the combined view. Removed the `links` import from search-helpers.ts entirely.

---

src/vault-mcp/search/search-helpers.ts
aliasunder: Fixed — added `__tests__/search-helpers.test.ts` with 37 assertions covering `isString`, `coerceToArray` (6 branches), `buildFtsMetadataText` (8 cases including object-skip and array-with-non-primitives), `mtimeToIso` (valid + invalid + fractional), `notePassesFilters` (7 cases: folder, tags, type, related, properties, AND combination), `buildSnippetFromChunkText` (4 cases), and `escapeLikeWildcards` (5 cases).

---

src/vault-mcp/search/search-queries.ts
aliasunder: Fixed — moved the FTS5 rank comment above the SQL template where the strategy is described.

---

src/vault-mcp/search/__tests__/fts-query.test.ts
aliasunder: Fixed — added both scenarios: `"meeting NEAR fence"` → `"meeting fence"` and `"NEAR"` alone → `'""'`.

---

src/vault-mcp/search/__tests__/rrf.test.ts
aliasunder: Fixed — replaced with exact `toEqual` assertion using the computed score `Number((1 / 11 + 0.05).toPrecision(4))`.

---

PR #289, opened 2026-07-08
fix: log error name and stack trace in error middleware, enable source maps

aliasunder:
### What

Unhandled errors in the Express error middleware only surfaced `err.message` — the error type and full call chain were invisible. This PR:

- Logs errors as `[ErrorName]: message` (e.g. `[TypeError]: bad type`) in both the error middleware and `describeError()` used by every background error site
- Logs `err.stack` for the full call chain
- Enables source maps in production (`tsconfig.json` + `Dockerfile`) so both the logger's `source` field and `err.stack` resolve to `.ts` lines instead of `.js`

### Changes

| File | Change |
|---|---|
| `src/vault-mcp/server.ts` | Error middleware logs `[${err.name}]: ${err.message}` + `err.stack` |
| `src/utils/describe-error.ts` | Now returns `[name]: message` for all `Error` instances — 20+ background error sites inherit this |
| `src/vault-mcp/vault-operations/daily-notes.ts` | `startsWith` guard updated for `[Error]:` prefix |
| `tsconfig.json` | `"sourceMap": true` |
| `Dockerfile` | `--enable-source-maps` in CMD |
| `src/vault-mcp/__tests__/server.test.ts` | 3 stack tests: present (exact assertion), absent session header (exact), undefined (new test) |
| `src/utils/__tests__/describe-error.test.ts` | Updated expectations for `[name]: message` format |
| `src/vault-mcp/search/__tests__/search-index.test.ts` | Updated `describeError` assertions |
| `src/vault-mcp/mcp-core/__tests__/tool-definitions.test.ts` | Updated `describeError` assertions |

### Verification

All 1576 tests pass across 50 test files.

PR #291, opened 2026-07-09
feat(docker)!: single OCI image — two-target build, fork absorption, image rename

aliasunder:
## What

Rewrites the Dockerfile as a **two-target build** and collapses the remote deployment to a **single OCI image**, eliminating Docker Compose as a user requirement — the remote stack now runs on any OCI runtime (`docker run`, Podman, nerdctl).

- **`local` target** (default stage, published as `:latest` / `:vX.Y.Z`) — tini + MCP server, behaviorally identical to today&#39;s `vault-mcp` image (verified via `docker inspect .Config` diff against the published image; only delta is `--enable-source-maps` from the not-yet-released #289)
- **`remote` target** (`:remote` / `:vX.Y.Z-remote`) — s6-overlay supervises obsidian-sync (`obsidian-headless` installed fresh via npm; the Alpine fork can&#39;t be a base since `onnxruntime-node` needs glibc) + the MCP server in one container
- **Fork absorption**: the obsidian-headless-sync-docker fork&#39;s `rootfs/` s6 scaffolding and `get-token.sh` now live in this repo, plus a new `svc-vault-mcp` longrun gated on `svc-obsidian-sync` (startup order only — same guarantee as the old `depends_on: service_healthy`)
- **Lightsail compose** simplifies to one service; volume names preserved so the cutover needs no data migration (vault, index, and Obsidian login state carry over; DEVICE_NAME unchanged → no new sync device)
- **Clean cutover, no dual-publish**: releases push only `vault-cortex` from this PR on — the legacy `vault-mcp` package freezes immediately (per-version GHCR download counts decompose to CI traffic + registry crawlers, with no residual attributable to third-party deployments; a frozen legacy stack keeps working). The rename is recorded in these release notes and the CHANGELOG rather than a README migration note
- **CI**: two build-push steps (shared-stage gha cache), trivy scans both targets on PRs and post-release, `--remove-orphans` retires the old two-service containers on first deploy. Carries the Test Deploy workflow (#293) adapted to this stack (`--target remote`, `vault-cortex:test`, `VAULT_CORTEX_TAG=test`)
- **Docs**: README variant explainer, ARCHITECTURE s6 startup section + diagram, DEPLOY/RECOVERY/AGENTS/llms-install updates, deploy/remote quickstart rewritten with a first-class `docker run` path

## Bug found during verification

Running the remote target against a real Obsidian Sync account surfaced a broken merge in the fork&#39;s 460f7d4 upstream sync: `init-setup-vault` passed a duplicated `sync-setup` token (`ob sync-setup sync-setup --vault …`), failing every fresh vault setup — this also affects the fork&#39;s published v0.1.3 image. Fixed in the absorbed copy (own commit, commented divergence).

## Verification

- Both targets build on arm64 (native M4) and amd64 (trivy-pr builds both in CI); `docker build .` with no `--target` produces a digest identical to `--target local`
- Local target: `/healthz` 200 + full MCP `initialize` handshake against a bind-mounted vault
- Remote target, real credentials (pull-only, throwaway device name): full s6 init chain (login → sync-setup), real vault synced into the volume, Docker health `healthy`, MCP handshake OK, both workload processes running as `obsidian` (only s6 machinery as root)
- No-token run exits with the init-check-auth remediation message (`S6_BEHAVIOUR_IF_STAGE2_FAILS=2`)
- PUID/PGID=1500 remap correctly rechowns `/vault`, `/data` (new, for the index volume), and `/home/obsidian`
- `get-token`, `ob`, `usermod` present in the remote image; npm/npx/corepack removed in both targets
- Lint, build, and all 1579 tests pass (includes the CLI template sync check and updated scaffold assertions)

BREAKING CHANGE: the published image is renamed `ghcr.io/aliasunder/vault-mcp` → `ghcr.io/aliasunder/vault-cortex` and now ships two targets (`:latest` = MCP server only; `:remote` = server + bundled Obsidian Sync under s6-overlay). The legacy `vault-mcp` package stops receiving tags at this release; already-published tags remain pullable indefinitely, frozen at the last pre-rename version. To migrate a local setup, point `image:` at `ghcr.io/aliasunder/vault-cortex:latest`; for a remote two-service setup, switch to the single-service `deploy/remote/docker-compose.yml` (`:remote` bundles sync) — the `.env` carries over unchanged.


.github/workflows/trivy.yml
```diff
@@ -45,12 +45,12 @@ jobs:
...
           persist-credentials: false
 
-      - name: Scan published image
+      - name: Scan published image (local target)
         uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # 0.36.0
         with:
           # Manifest list resolves per-runner (amd64 here); the Debian OS
           # package set is identical across arches, so amd64-only is fine.
-          image-ref: ghcr.io/${{ vars.GHCR_USER }}/vault-mcp:latest
+          image-ref: ghcr.io/${{ vars.GHCR_USER }}/vault-cortex:latest
```
aliasunder, line 74: **[D1]** Guaranteed workflow failure + red badge on the merge commit — `ghcr.io/aliasunder/vault-cortex` does not exist yet (verified: GHCR package API returns 404). `trivy-published` runs on `push: branches: [main]`, so the run triggered by merging this PR will fail pulling `:latest` (and `:remote` below) before any release can publish the image, and the README Trivy badge goes red until the first post-merge release.

Options: cut a release immediately after merging (simplest — the failure window closes at the first `deploy.yml` run), or gate the scan steps until first publish (e.g. a `docker manifest inspect` pre-check step, removed later). Flagged: needs design decision (process vs. temporary guard).

---
*🔍 ship-check · pr-review · claude-fable-5*

.github/workflows/ghcr-cleanup.yml
```diff
@@ -23,6 +23,16 @@ jobs:
       packages: write
     steps:
       - name: Delete untagged GHCR images
+        uses: dataaxiom/ghcr-cleanup-action@d52806a0dc70b430571a37da1fde39733ffd640f # v1.2.2
+        with:
+          package: vault-cortex
```
aliasunder, line 28: **[D1]** Same first-release timing gap as the trivy comment: the `vault-cortex` GHCR package doesn't exist yet (API 404), and this workflow runs weekly (Sunday 03:00 UTC). If the cron fires before the first post-merge release pushes the image, this step fails on a missing package. Self-heals at the first release; only worth a guard if the merge→release gap could span a Sunday. Flagged: needs design decision (same resolution as the trivy timing comment).

---
*🔍 ship-check · pr-review · claude-fable-5*

rootfs/etc/s6-overlay/scripts/init-check-auth
```diff
@@ -0,0 +1,11 @@
+#!/command/with-contenv sh
+# Validate that OBSIDIAN_AUTH_TOKEN is set before proceeding.
+set -e
+
+if [ -z "$OBSIDIAN_AUTH_TOKEN" ]; then
```
aliasunder, line 5: **[D1]** `VAULT_NAME` has no fail-fast guard on the newly first-class `docker run` path. Compose enforces it via `${VAULT_NAME:?…}`, but `docker run --env-file .env` (now a documented first-class path in deploy/remote/README.md) does no `:?` interpolation — an empty `VAULT_NAME=` slips through, `init-setup-vault` silently skips `ob sync-setup` (its whole body is gated on `[ -n "$VAULT_NAME" ]`), and `svc-obsidian-sync` then runs `ob sync --continuous` against an unconfigured vault — a confusing crash-loop with the MCP server gated behind it, instead of the crisp remediation message the token gets.

Suggested fix (~4 lines): extend this script to also require `VAULT_NAME`, mirroring the token check:

```sh
if [ -z "$VAULT_NAME" ]; then
  echo "[obsidian-sync] ERROR: VAULT_NAME is not set." >&2
  echo "[obsidian-sync] Set VAULT_NAME to your Obsidian vault name (case-sensitive)." >&2
  exit 1
fi
```

There's no legitimate empty-`VAULT_NAME` flow for this image (a no-sync deployment is the `:latest` target), so the fork-inherited `-n` gate in init-setup-vault has nothing to preserve. Would fix (trivial).

---
*🔍 ship-check · pr-review · claude-fable-5*

Dockerfile
```diff
@@ -55,7 +65,95 @@ COPY --from=deps  /app/node_modules ./node_modules
...
+         arm64) echo aarch64;; \
+         *) echo "Unsupported architecture: ${TARGETARCH}" >&2; exit 1;; \
+       esac)" \
+    && S6_BASE="https://github.com/just-containers/s6-overlay/releases/download/v${S6_OVERLAY_VERSION}" \
+    && wget -qO /tmp/s6-overlay-noarch.tar.xz "${S6_BASE}/s6-overlay-noarch.tar.xz" \
+    && wget -qO /tmp/s6-overlay-noarch.tar.xz.sha256 "${S6_BASE}/s6-overlay-noarch.tar.xz.sha256" \
+    && wget -qO /tmp/s6-overlay-${S6_ARCH}.tar.xz "${S6_BASE}/s6-overlay-${S6_ARCH}.tar.xz" \
+    && wget -qO /tmp/s6-overlay-${S6_ARCH}.tar.xz.sha256 "${S6_BASE}/s6-overlay-${S6_ARCH}.tar.xz.sha256" \
+    && cd /tmp \
+    && sha256sum -c s6-overlay-noarch.tar.xz.sha256 \
```
aliasunder: **[D4]** The `.sha256` files are downloaded from the same GitHub release as the tarballs, so `sha256sum -c` verifies transfer integrity only — a compromised or re-uploaded release asset ships matching checksums. This is an internal-consistency gap rather than generic advice: the image digest-pins its base specifically against tag-mutation supply-chain risk (per the comment above and the ARCHITECTURE hardening table), and GitHub release assets are mutable the same way tags are.

Suggested fix: embed the known digests as constants (noarch + x86_64 + aarch64 — the arch case already exists) and drop the two `.sha256` downloads; refresh them when `S6_OVERLAY_VERSION` bumps. Would fix (small; needs the three published hashes for v3.2.2.0).

---
*🔍 ship-check · pr-review · claude-fable-5*

.github/workflows/trivy.yml
```diff
@@ -109,3 +154,10 @@ jobs:
         with:
           sarif_file: trivy-results.sarif
           category: trivy-pr-image
+
+      - name: Upload remote SARIF to the Security tab
+        if: always() && github.event.pull_request.head.repo.full_name == github.repository
```
aliasunder: **[D4]** Minor CI noise: `trivy-results-remote.sarif` won't exist when the job fails before the remote scan runs — the local scan gates with `exit-code: "1"`, and a remote build failure skips the scan — so this `if: always()` upload then fails with a missing-file error layered on top of the real failure. Same applies to the scheduled job's remote upload (line 82).

Suggested guard: `if: always() && hashFiles('trivy-results-remote.sarif') != '' && github.event.pull_request.head.repo.full_name == github.repository`. Would fix (trivial).

---
*🔍 ship-check · pr-review · claude-fable-5*

DEPLOY.md
```diff
@@ -302,41 +302,41 @@ ssh -i ~/.ssh/vault-cortex ubuntu@<lightsailIp>
 
 ### Tailing logs
 
-Both containers write structured JSON to stdout/stderr, captured by Docker's `json-file` log driver (10MB per file, 3 rotated files — ~30MB retained per container).
+The container writes to stdout/stderr, captured by Docker's `json-file` log driver (10MB per file, 3 rotated files — ~30MB retained). One stream carries both processes: sync lines are prefixed `[obsidian-sync]`; MCP server lines are structured JSON.
```
aliasunder: **[D1]** Doc accuracy: "sync lines are prefixed `[obsidian-sync]`" overstates what the rootfs actually does. The prefix appears only on the init-chain scripts' own `echo` lines and the one startup banner in `svc-obsidian-sync/run` — the `ob sync --continuous` process is `exec`'d directly with no s6 log pipeline, so its ongoing output is **unprefixed**. s6-overlay does no prefixing by default. Consequently the documented workflow `docker logs vault-cortex 2>&1 | grep obsidian-sync` (line 318) surfaces only init/banner lines, not sync activity. Same claim in deploy/remote/README.md:189 ("s6 prefixes sync lines with [obsidian-sync]").

Minimal fix: reword both spots — e.g. "init/setup lines are prefixed `[obsidian-sync]`; the sync process's own output is plain text (vs. the MCP server's structured JSON — filter sync activity with `grep -v '^{'`)". Actually prefixing `ob`'s output would need an s6 log service or pipeline — flagged: needs design decision if you want the grep workflow to work as written; the doc reword itself is a would-fix.

---
*🔍 ship-check · pr-review · claude-fable-5*

aliasunder (commented):
## Phase 1: PR Review

6 findings across 6 files (5 inline, 1 below — file not in the diff). Dimensions: correctness, security/performance, feature-surface docs, stale paths. TDQS: N/A (no tool description changes). Stale-path sweep of AGENTS/ARCHITECTURE/README/DEPLOY/RECOVERY/llms-install/wiki.json/CLI came back clean apart from the item below. The fork's duplicated `sync-setup` token bug was independently verified against the fork's current `init-setup-vault` (`set -- sync-setup --vault …`) — the fix here is correct.

### [D6] SECURITY.md — stale deployment scope and container-hardening claims (not in this diff, so no inline comment)

- Line ~20: "**Docker Compose** — two long-running containers on Lightsail sharing a `/vault` volume (UID 1000)" — the Lightsail stack is now one `vault-cortex:remote` container.
- Lines ~95–96: "Non-root user (`USER node`, UID 1000)" and "PID 1 init (`tini`)" — accurate for the `:latest` target only. In the `:remote` image, s6's `/init` runs as PID 1 **as root** and the two services drop to UID 1000 via `s6-setuidgid`; there is no `USER` directive. ARCHITECTURE.md's hardening table was updated with exactly this nuance, so SECURITY.md now contradicts it — and a security doc overstating "non-root PID 1" is the kind of claim OSS adopters check. Would fix (small doc edit mirroring the ARCHITECTURE table wording).

**Verdict**: ship-with-minor-fixes

---
*🔍 ship-check · pr-review · claude-fable-5*

rootfs/etc/s6-overlay/scripts/init-setup-vault
```diff
@@ -0,0 +1,77 @@
...
+    fi
+    exit 1
+  fi
+fi
+
+# ---------------------------------------------------------------------------
+# Apply optional sync config
+# ---------------------------------------------------------------------------
+if [ -n "$DEVICE_NAME" ]; then
+  s6-setuidgid obsidian ob sync-config --device-name "$DEVICE_NAME" 2>/dev/null || true
```
aliasunder: **[error-handling]** Each of the six optional sync-config blocks (lines 55–77) swallows failures silently — `2>/dev/null` discards the CLI's stderr *and* `|| true` masks the exit code. If a user sets a bad value (e.g. a typo'd `CONFLICT_STRATEGY` via `docker run`, where compose defaults don't apply), their requested config silently never applies and `docker logs` contains zero trace. Non-fatal is the right call for optional config, but it should warn. A small helper keeps non-fatal behavior, surfaces the failed flag + stderr in the log, and collapses the six near-identical blocks:

```sh
# Optional config — a failure logs a warning but does not stop the container.
apply_sync_config() {
  if [ -n "$2" ] && ! s6-setuidgid obsidian ob sync-config "$1" "$2"; then
    echo "[obsidian-sync] WARN: ob sync-config $1 failed — continuing with the previous setting." >&2
  fi
}

apply_sync_config --device-name "$DEVICE_NAME"
apply_sync_config --conflict-strategy "$CONFLICT_STRATEGY"
apply_sync_config --excluded-folders "$EXCLUDED_FOLDERS"
apply_sync_config --file-types "$FILE_TYPES"
apply_sync_config --mode "$SYNC_MODE"
apply_sync_config --configs "$SYNC_CONFIGS"
```

(Note: `set -e` doesn't fire inside an `if !` condition, so the helper stays non-fatal.)

---
*🔍 ship-check · code-quality · claude-fable-5*

cli/src/init.ts
```diff
@@ -57,7 +57,7 @@ const askMode = async (prompts: Prompts): Promise<Mode> => {
 }
 
 const GET_TOKEN_COMMAND = `docker run --rm -it --entrypoint get-token \\
-  ${OBSIDIAN_SYNC_IMAGE}`
+  ${GET_TOKEN_IMAGE}`
 
 /**
  * Offers to run the obsidian-headless-sync get-token flow in this terminal.
```
aliasunder: **[comments]** Stale image reference — the docstring still says "obsidian-headless-sync get-token flow", but that image is retired; the flow now runs the `vault-cortex:remote` image's own `get-token` entrypoint (`GET_TOKEN_IMAGE`, two lines up).

```suggestion
 * Offers to run the image's get-token flow in this terminal.
```

---
*🔍 ship-check · code-quality · claude-fable-5*

aliasunder (commented):
## Phase 2: Code Quality

3 findings across 3 files (error handling, comments, naming). Everything else in the changed surface — Dockerfile stage structure, workflow comments, compose restructuring, CLI renames — reads clean against AGENTS.md conventions.

---
*🔍 ship-check · code-quality · claude-fable-5*

get-token.sh
```diff
@@ -0,0 +1,62 @@
...
+
+for candidate in $CANDIDATES; do
+  if [ -f "$candidate" ]; then
+    TOKEN=$(cat "$candidate")
+    break
+  fi
+done
+
+if [ -z "$TOKEN" ]; then
+  FOUND=$(find "$HOME" -path "*obsidian-headless*" -name "auth_token" 2>/dev/null | head -1)
```
aliasunder: **[naming]** `FOUND` doesn't say what the value is — it holds the path to the located auth-token file. Name it for its contents:

```sh
TOKEN_FILE=$(find "$HOME" -path "*obsidian-headless*" -name "auth_token" 2>/dev/null | head -1)
if [ -n "$TOKEN_FILE" ]; then
  TOKEN=$(cat "$TOKEN_FILE")
fi
```

---
*🔍 ship-check · code-quality · claude-fable-5*

cli/src/__tests__/scaffold.test.ts
```diff
@@ -25,22 +25,20 @@ describe("buildFilesToWrite", () => {
...
     const envFile = files.find((file) => file.name === ".env")
     expect(composeFile?.content).toContain(
-      "ghcr.io/aliasunder/vault-mcp:latest",
+      "ghcr.io/aliasunder/vault-cortex:latest",
     )
     expect(envFile?.content).toBe("MCP_AUTH_TOKEN=abc\n")
   })
 
-  it("plans the two-service remote compose with the sync image and no init-config-perms", () => {
+  it("plans the single-service remote compose with the :remote image and no separate sync image", () => {
```
aliasunder, line 33: **[two-bar]** Test name claims "single-service" but no assertion checks the service count.

Mutation-verified: appending a second service with a fresh image name to both remote compose copies leaves the entire suite green — including this test. The `not.toContain("obsidian-headless-sync-docker")` guardrail only blocks the *retired* image from returning, not a second service in general.

Either assert single-service-ness (e.g. exactly one `image:` line in the compose content: `expect(composeContent?.match(/^\s+image:/gm)).toHaveLength(1)`), or rename the test to match what it asserts (":remote image, no retired sync image").

---
*🔍 ship-check · test-audit · claude-fable-5*

aliasunder (commented):
## Phase 3: Test Audit

1 test quality finding, 2 coverage gaps across 3 files. All 110 CLI tests pass at the PR head; each finding below was verified by mutation (the described breakage leaves the full suite green).

Audited: `cli/src/__tests__/scaffold.test.ts` (all 12 `it()` blocks) plus coverage-gap analysis on `cli/src/docker.ts`, `cli/src/env.ts`, `cli/src/init.ts`, `scripts/dev.ts`. The `writeFiles`/`readEnvPort`/permissions tests are sound (exact result-array assertions, whole-content `toBe`, throwing-resolver sentinel). Removed assertions in the scaffold diff (`obsidian-sync` service, `init-config-perms`, retired image names) are obsolete under the single-image architecture — no coverage regression. `scripts/dev.ts` is an ops helper with no vitest precedent under `scripts/` — not flagged. `init.ts`'s `GET_TOKEN_COMMAND` is covered transitively by the `docker.ts` drift-test suggestion below.

---
*🔍 ship-check · test-audit · claude-fable-5*

cli/src/docker.ts
```diff
@@ -11,8 +11,7 @@ export type DockerRunner = {
 }
 
 /** The image whose `get-token` entrypoint issues Obsidian Sync auth tokens. */
-export const OBSIDIAN_SYNC_IMAGE =
-  "ghcr.io/aliasunder/obsidian-headless-sync-docker:latest"
+export const GET_TOKEN_IMAGE = "ghcr.io/aliasunder/vault-cortex:remote"
```
aliasunder, line 14: **[coverage gap]** Nothing binds `GET_TOKEN_IMAGE` to the image the remote compose actually pulls.

Mutation-verified: reverting this constant to the retired `obsidian-headless-sync-docker` image leaves all 110 CLI tests green. This constant feeds the get-token command shown to every remote user (`init.ts` `GET_TOKEN_COMMAND` → `prompts.note`), so silent drift here means users are told to pull the wrong image — exactly the cross-file sweep this PR had to do by hand.

A drift test in the `templates.test.ts` style would pin it: `expect(readComposeTemplate("remote")).toContain(\`image: ${GET_TOKEN_IMAGE}\`)` — a future rename/tag change that updates the compose but not the CLI (or vice versa) then fails a test.

---
*🔍 ship-check · test-audit · claude-fable-5*

cli/src/env.ts
```diff
@@ -151,7 +151,7 @@ VAULT_PASSWORD=${answers.vaultPassword}`
       ? `# Obsidian Sync auth token — FILL THIS IN before docker compose up.
 # Generate once with:
 #   docker run --rm -it --entrypoint get-token \\
-#     ghcr.io/aliasunder/obsidian-headless-sync-docker:latest`
+#     ghcr.io/aliasunder/vault-cortex:remote`
```
aliasunder: **[coverage gap]** The get-token image in generated `.env` guidance is an independent string literal — no test pins it.

Mutation-verified: reverting this line to the retired `obsidian-headless-sync-docker` image leaves the suite green. The existing test (`env.test.ts`: "writes an empty OBSIDIAN_AUTH_TOKEN line with a fill-this-in warning…") asserts only `toContain("get-token")`, which passes with either image — so this literal can drift from `GET_TOKEN_IMAGE` in `docker.ts` silently.

Strengthen that test to assert the generated env contains `GET_TOKEN_IMAGE` (import it from `docker.js`), tying the duplicate literal to the constant.

---
*🔍 ship-check · test-audit · claude-fable-5*

deploy/remote/README.md
```diff
@@ -60,9 +67,26 @@ cp .env.example .env
...
+vault size. The MCP server starts once sync is running and builds its search
+index as files arrive.
+
+### docker run (no Compose)
+
+The same `.env` file works with any OCI runtime — swap `docker` for `podman`
+or `nerdctl` as needed:
+
+```bash
+docker run -d --name vault-cortex \
```
aliasunder, line 86: **[D5]** The `docker run` path silently loses the compose-supplied defaults — most damagingly `DEVICE_NAME`, so the initial Sync device registers under a random container-ID hostname.

Line 77 claims: "The same `.env` file works with any OCI runtime". But the compose file supplies defaults the `.env.example` ships commented out: `hostname: ${DEVICE_NAME:-vault-cortex}` + `DEVICE_NAME: ${DEVICE_NAME:-vault-cortex}`, `CONFLICT_STRATEGY:-merge`, `SYNC_MODE:-bidirectional`, `LOG_DIR:-/data/logs`. This `docker run` one-liner passes none of them. With `DEVICE_NAME` unset, `init-setup-vault` omits `--device-name`, and per that script's own comment: "Without this the CLI registers the first device under the container hostname … so every container recreation spawns a new Obsidian Sync device." `docker run` without `--hostname` gives a random 12-hex-char container-ID hostname — so the documented one-liner registers a garbage-named device and re-registers a new one on every recreation. `CONFLICT_STRATEGY`/`SYNC_MODE` fall back to `ob` defaults instead of compose's `merge`/`bidirectional`, and file logging is silently off (`LOG_DIR` unset).

Suggested fix: add `--hostname vault-cortex \` (or `-e DEVICE_NAME=vault-cortex \`) to the snippet, and either add the other three as `-e` lines or tell docker-run users to uncomment `DEVICE_NAME`/`CONFLICT_STRATEGY`/`SYNC_MODE`/`LOG_DIR` in `.env` since compose defaults don't apply outside compose.

*Would fix (trivial, high confidence)*

---
*🔍 ship-check · bug-check · claude-fable-5*

rootfs/etc/s6-overlay/s6-rc.d/svc-vault-mcp/run
```diff
@@ -0,0 +1,13 @@
+#!/command/with-contenv sh
+# MCP server longrun supervised by s6.
+# The dependency on svc-obsidian-sync gates STARTUP ORDER only — s6-rc does
+# not restart-couple longruns — which matches what the old two-container
+# Compose stack guaranteed (depends_on: service_healthy = "sync process has
```
aliasunder: **[D1]** "matches what the old two-container Compose stack guaranteed" overclaims — the new guarantee is strictly weaker on fresh volumes.

Old stack: `depends_on: condition: service_healthy` blocked vault-mcp's *start* until obsidian-sync's first healthcheck probe **passed**, and Docker runs the first probe only after `interval` (30s) — so the MCP server always started ≥~30s after sync began, and the old ARCHITECTURE text leaned on exactly that: "sync is up and has had its `start_period` to land files … **Combined with** the memory-write shrink guard, that's enough to prevent the fresh-volume clobber."

New stack: an s6-rc longrun without `notification-fd` is considered up the moment its run process spawns, so `svc-vault-mcp` starts milliseconds after `ob sync` execs. The ~30s head start for the first sync is gone, and the memory-bootstrap race window on a fresh volume is now bounded only by the shrink guard — one of the two legs of the old protection. Same claim appears at ARCHITECTURE.md:516 ("the same guarantee") and in the PR body.

Two parts:
1. *Would fix (trivial, medium confidence):* reword here + ARCHITECTURE.md:516 — "a similar but slightly weaker guarantee: the old stack also waited out the first healthcheck probe (~30s), the s6 dependency does not."
2. *Flagged: needs design decision* — whether to restore a head start (s6 readiness `notification-fd` on svc-obsidian-sync, or a short first-boot delay) or accept the shrink guard as sufficient.

---
*🔍 ship-check · bug-check · claude-fable-5*

rootfs/etc/s6-overlay/scripts/init-setup-user
```diff
@@ -0,0 +1,40 @@
...
+  usermod -o -u "$PUID" obsidian
+  ID_CHANGED=true
+fi
+
+# Ensure the vault and index directories exist.
+# /data is vault-cortex's index volume (INDEX_DB_PATH) — it must follow
+# PUID/PGID alongside /vault so svc-vault-mcp can write the search index.
+mkdir -p "$VAULT_PATH" /data
+
+if [ "$ID_CHANGED" = true ]; then
```
aliasunder: **[D6]** PUID remap detection keys off the image-baked passwd entry, so reverting PUID back to 1000 strands existing volume contents at the old UID.

`ID_CHANGED` compares `PUID` against `id -u obsidian`, which on every container **recreation** is the image default 1000 — volume history never enters the check. Scenario: run with `PUID=1500` (recursive chown makes all volume contents 1500) → later remove `PUID` (defaults back to 1000) → new container sees `1000 == 1000` → `ID_CHANGED=false` → only the top-level `chown` of `/vault`, `/data`, `/home/obsidian` runs → every nested file stays owned 1500, and both services (now running as 1000) fail to write existing notes/index files with no hint from this script's "Running as UID=1000" log.

Fix shape: derive the "changed" signal from actual volume ownership (e.g. `stat -c %u "$VAULT_PATH"` vs `$PUID`, falling back to top-level chown when they match) instead of the passwd entry — or always `chown -R`, which penalizes large vaults on every boot. Behavior is absorbed from the fork (pre-existing there, new to this repo).

*Flagged: needs design decision (high confidence in the mechanism, non-trivial fix tradeoffs)*

---
*🔍 ship-check · bug-check · claude-fable-5*

deploy/remote/.env.example
```diff
@@ -14,7 +14,7 @@ PUBLIC_URL=
 
 # Obsidian Sync auth token. Generate once with:
 #   docker run --rm -it --entrypoint get-token \
-#     ghcr.io/aliasunder/obsidian-headless-sync-docker:latest
+#     ghcr.io/aliasunder/vault-cortex:remote
```
aliasunder, line 17: **[D5]** Pre-existing gap this PR's compose rewrite makes visible: this file documents `WINDOWS_MODE` (block near line 47, "Set to true when your vault is on a Windows drive"), but neither `deploy/remote/docker-compose.yml` nor `cli/templates/remote/docker-compose.yml` passes `WINDOWS_MODE` to the container — setting it per these instructions is a silent no-op. (The root Lightsail compose *does* pass it; only the remote quickstart surfaces are missing it.) The CLI's `REMOTE_OPTIONAL_BLOCK` in `cli/src/env.ts` already omits the var, so the two files AGENTS.md says must mirror each other also disagree.

Either direction is a small fix: pass `WINDOWS_MODE: ${WINDOWS_MODE:-false}` through both remote compose files (matching the root compose), or drop the block from this file (named-volume remote deployments never hit the Docker Desktop/WSL2 bind-mount problem, and dropping matches the CLI env builder). Dropping looks more correct.

*Would fix (trivial, high confidence) — pre-existing gap*

---
*🔍 ship-check · bug-check · claude-fable-5*

README.md
```diff
@@ -305,6 +305,11 @@ See [ARCHITECTURE.md → Auth](./ARCHITECTURE.md#auth-oauth-21--defense-in-depth
 | **Remote**    | VPS + Obsidian Sync, access from anywhere          | [`deploy/remote/`](./deploy/remote/) |
 | **AWS (SST)** | Full IaC: Lightsail + API Gateway + Lambda + CI/CD | [`DEPLOY.md`](./DEPLOY.md)           |
 
+Both paths run the same image, `ghcr.io/aliasunder/vault-cortex` — `:latest` is the MCP server alone (local), `:remote` bundles Obsidian Sync in the same container under [s6-overlay](https://github.com/just-containers/s6-overlay) supervision. One container means any OCI runtime works: `docker run`, Podman, nerdctl — Docker Compose is optional.
+
+> [!NOTE]
+> **Migrating from `vault-mcp`?** The image was renamed to `ghcr.io/aliasunder/vault-cortex`. The old `vault-mcp` name keeps receiving release tags until **2026-08-08** — update the `image:` line in your compose file before then. Already-published `vault-mcp` tags remain pullable indefinitely; they just stop updating.
```
aliasunder: **[D1]** The migration note is complete for local users but misleading for remote-quickstart users.

"update the `image:` line in your compose file before then" — a user on the old two-service remote compose (`vault-mcp` + `obsidian-headless-sync-docker`) who follows this literally swaps `vault-mcp` → `vault-cortex:latest` and keeps the separate `obsidian-sync` service pointing at the fork image — which, per this PR's own Acknowledgments change, now "lives in this repo" and stops receiving lockstep releases. Their sync image silently freezes while everything appears to keep working.

Suggested fix: one added sentence — remote (Obsidian Sync) users should replace their compose file with the new single-service `:remote` one (volume names are unchanged, so the vault, index, and login state carry over) and run `docker compose up -d --remove-orphans` to retire the old sync container.

*Would fix (trivial, medium confidence)*

---
*🔍 ship-check · bug-check · claude-fable-5*

aliasunder (commented):
## Phase 4: Bug Check

8 findings across 8 files (6 would-fix, 3 flagged — one finding has a dual disposition).
Confidence: 4 high, 2 medium, 2 low.
By dimension: description-vs-implementation 4, behavioral consistency 3, input validation/boundary 1.

Verified as non-issues (checked, not skipped): transformers model cache is cwd-relative (`/app` in both eras — HOME change is inert); `ca-certificates` survives the `apt-get autoremove` (explicitly installed); the `start_period`/retries healthcheck comments match Docker semantics; `docker build .` with no `--target` correctly produces the `local` stage (kept last); volume-name preservation checked across all three remote compose files; `--remove-orphans` correctly retires the old two-service containers (same Compose project name).

---
*🔍 ship-check · bug-check · claude-fable-5*

cli/src/init.ts
```diff
@@ -57,7 +57,7 @@ const askMode = async (prompts: Prompts): Promise<Mode> => {
 }
 
 const GET_TOKEN_COMMAND = `docker run --rm -it --entrypoint get-token \\
-  ${OBSIDIAN_SYNC_IMAGE}`
+  ${GET_TOKEN_IMAGE}`
```
aliasunder, line 60: **[D1]** Two more stale comments in this file (beyond the line-63 docstring code-quality already raised) — neither line is in a diff hunk, so anchoring here:

- Line 356: "write the three-service compose + .env" — the remote compose this flow writes is now **single**-service (and was two-service before this PR; the count was never right).
- Line 433: "Without the sync token the stack can't start (obsidian-sync exits)" — there is no `obsidian-sync` service anymore; the failure mode is now the whole container exiting when `init-check-auth` fails (`S6_BEHAVIOUR_IF_STAGE2_FAILS=2`).

*Would fix (trivial, high confidence) — line 356 pre-existing, line 433 stale as of this PR*

---
*🔍 ship-check · bug-check · claude-fable-5*

.env.example
```diff
@@ -37,6 +37,10 @@ GHCR_TOKEN=
 
 # Docker
 GHCR_USER=
+# Image tag the Lightsail compose pulls (default: remote — the target that
+# bundles Obsidian Sync). Pin to a versioned tag (e.g. v0.27.0-remote) to
+# hold a release.
+# VAULT_CORTEX_TAG=remote
```
aliasunder, line 43: **[D5]** Silent knob retirement: an existing Lightsail `.env` pinning `VAULT_MCP_TAG` (the old compose's `vault-mcp:${VAULT_MCP_TAG:-latest}`) is silently ignored after the cutover — compose doesn't warn on unreferenced env vars, so a previously pinned deployment starts pulling `:remote` (latest) with no signal. Exactly the silent-misconfiguration class a "no data migration needed" cutover should call out.

Suggested fix: one line in this comment block — "(renamed from VAULT_MCP_TAG; re-pin under the new name if you had it set)" — and/or a mention in DEPLOY.md.

*Would fix (trivial, low confidence of real-world impact — reference deployment only)*

---
*🔍 ship-check · bug-check · claude-fable-5*

.github/workflows/deploy.yml
```diff
@@ -155,6 +165,26 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+      - name: Build and push image (remote target)
```
aliasunder: **[D1]** Both build-push steps push different `index:org.opencontainers.image.description` annotations to the **same** GHCR package (`vault-cortex`), and this remote step runs second. If GHCR displays the package-page description from the most recently pushed version (the commonly observed behavior — I could not verify it from here), the page will show the remote-flavored text, contradicting the local step's comment that its annotation is what mirrors `server.json` on the package page.

If confirmed, the fix is free: swap the two steps (remote first, local second) so the `:latest`-matching description wins. Worth a one-time check on the package page after the first release either way.

*Flagged: uncertain diagnosis (low confidence — GHCR last-push-wins behavior unverified)*

---
*🔍 ship-check · bug-check · claude-fable-5*

scripts/dev.ts
```diff
@@ -173,9 +175,9 @@
     )
     run(`scp ${id} ${sshOpts} ${ENV_PATH} ubuntu@${ip}:/opt/vault-cortex/.env`)
     run(
-      `ssh ${id} ${sshOpts} ubuntu@${ip} 'cd /opt/vault-cortex && docker compose pull && docker compose up -d && docker image prune -f'`,
+      `ssh ${id} ${sshOpts} ubuntu@${ip} 'cd /opt/vault-cortex && docker compose pull && docker compose up -d --remove-orphans && docker image prune -f'`,
     )
-    console.log(`✓ vault-mcp deployed to ${ip}:8000`)
+    console.log(`✓ vault-cortex deployed to ${ip}:8000`)
```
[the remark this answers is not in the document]
aliasunder: Fixed in b86f7b8 — the deploy success log no longer prints the instance IP at all (stronger than the existing `::add-mask::` mitigation, and consistent with the repo's posture of keeping the IP out of SST outputs).

---

scripts/dev.ts
```diff
@@ -173,9 +175,9 @@ switch (sub) {
     )
     run(`scp ${id} ${sshOpts} ${ENV_PATH} ubuntu@${ip}:/opt/vault-cortex/.env`)
     run(
-      `ssh ${id} ${sshOpts} ubuntu@${ip} 'cd /opt/vault-cortex && docker compose pull && docker compose up -d && docker image prune -f'`,
+      `ssh ${id} ${sshOpts} ubuntu@${ip} 'cd /opt/vault-cortex && docker compose pull && docker compose up -d --remove-orphans && docker image prune -f'`,
     )
```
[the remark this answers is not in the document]
aliasunder: Fixed in b86f7b8 — `lightsail:up` now runs `docker compose up -d --remove-orphans --wait --wait-timeout 300`, matching deploy.yml.

---

cli/src/__tests__/scaffold.test.ts
```diff
@@ -25,22 +25,20 @@ describe("buildFilesToWrite", () => {
...
     )?.content
 
-    expect(composeContent).toContain("obsidian-sync")
-    expect(composeContent).toContain(
-      "ghcr.io/aliasunder/obsidian-headless-sync-docker:latest",
-    )
-    expect(composeContent).not.toContain("init-config-perms")
+    expect(composeContent).toContain("ghcr.io/aliasunder/vault-cortex:remote")
+    expect(composeContent).toContain("OBSIDIAN_AUTH_TOKEN")
+    expect(composeContent).not.toContain("obsidian-headless-sync-docker")
```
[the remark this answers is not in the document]
aliasunder: Fixed in b86f7b8 — the remote scaffold test now also asserts `MCP_AUTH_TOKEN` is present and that exactly one `image:` line exists (single-service check via regex count, no YAML parser needed).

---

ARCHITECTURE.md
```diff
@@ -477,47 +488,54 @@ both. Existing JWTs signed with the old key become invalid immediately.
...
+3. **`svc-vault-mcp`** — MCP server. Drops to the same `obsidian` user, so
+   both processes read/write the shared `/vault` volume. On startup: builds
+   the FTS5 search index, bootstraps memory templates if the memory folder
+   doesn't exist and `MEMORY_ENABLED` is not `false`, then starts the file
+   watcher.
+
+`svc-vault-mcp` declares `svc-obsidian-sync` in its `dependencies.d`, so the
+MCP server starts only after login and vault setup completed and the sync
+process has spawned — the same guarantee the previous two-container stack's
+`depends_on: service_healthy` (pgrep-based) provided. Note the guarantee is
```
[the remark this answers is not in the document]
aliasunder: Fixed in b86f7b8 — reworded to "after login and vault setup have completed" (the sentence was also revised for accuracy per a separate finding on the same paragraph).

---

rootfs/etc/s6-overlay/scripts/init-obsidian-login
```diff
@@ -0,0 +1,11 @@
+#!/command/with-contenv sh
+# Authenticate with Obsidian using the provided auth token.
+set -e
+
+echo "[obsidian-sync] Logging in to Obsidian..."
+if ! s6-setuidgid obsidian ob login; then
+  echo "[obsidian-sync] ERROR: ob login failed." >&2
+  exit 1
+fi
```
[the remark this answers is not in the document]
aliasunder: False positive — verified against the actual CLI in the built image: `ob login --help` accepts only `--email/--password/--mfa` (passing `--config-dir` would error), and `ob sync-setup --help` shows `--config-dir <name>` is the in-vault Obsidian config directory name (default `.obsidian`) used for sync configuration — not the auth-token storage location. The token lives under `$HOME/.config/obsidian-headless` regardless of CONFIG_DIR_NAME, so there is no mismatch.

---

.github/workflows/trivy.yml
aliasunder: Fixed in b86f7b8 — `trivy-published` now gates both scan steps on `docker manifest inspect` succeeding for the corresponding tag (with a `::notice::` when skipping), so the merge → first-release window produces a green run instead of a red badge. We also plan to cut the release immediately after merging, which closes the window in practice.

---

.github/workflows/ghcr-cleanup.yml
aliasunder: Design decision, as flagged: the resolution is releasing immediately after merge (same as the trivy thread), which closes the gap well before the next Sunday cron. No guard added — the failure mode is a soft red X on a weekly housekeeping run that self-heals at first release, and a manifest-existence guard here would outlive its usefulness. The trivy workflow got the guard because it also fires on every push to main and drives the README badge.

---

rootfs/etc/s6-overlay/scripts/init-check-auth
aliasunder: Fixed in b86f7b8 — `init-setup-vault` now logs a loud two-line WARNING to stderr when VAULT_NAME is unset (naming the crash-loop consequence). Deliberately a warning rather than a hard fail in `init-check-auth`: a container whose config volume already holds a completed sync setup legitimately restarts without VAULT_NAME, and hard-failing would break that path.

---

Dockerfile
aliasunder: Fixed in b86f7b8 — the three tarball digests (noarch, x86_64, aarch64) are now embedded as ARGs and verified via `echo "<digest>  <file>" | sha256sum -c`; the same-origin `.sha256` downloads are gone. Comment documents refreshing them on version bumps.

---

.github/workflows/trivy.yml
aliasunder: Fixed in b86f7b8 — `hashFiles(...) != ''` guards added to all four SARIF uploads (PR local + remote, published local + remote), so a gated or failed scan no longer layers a missing-file error on top of the real outcome.

---

DEPLOY.md
aliasunder: Fixed in b86f7b8 — both DEPLOY.md and deploy/remote/README.md now say the `[obsidian-sync]` prefix covers only init-chain/lifecycle lines while the continuous sync output is plain unprefixed text, and the filter recipe is now `docker logs vault-cortex 2>&1 | grep -v '^{'` (everything non-JSON is the sync side), which actually surfaces sync activity.

---

rootfs/etc/s6-overlay/scripts/init-setup-vault
aliasunder: Fixed in b86f7b8 — the six blocks collapsed into an `apply_sync_config` helper: still non-fatal, but stderr flows to the log and a failure prints `WARNING: ob sync-config <flag> '<value>' failed — continuing without it`. Verified in a real-credentials boot (DEVICE_NAME + SYNC_MODE applied cleanly, no spurious warnings).

---

cli/src/init.ts
aliasunder: Fixed in b86f7b8 — docstring now reads "the vault-cortex image's get-token flow".

---

get-token.sh
aliasunder: Fixed in b86f7b8 — `FOUND` renamed to `TOKEN_FILE` (a deliberate divergence from the fork copy, which is being absorbed and archived).

---

cli/src/__tests__/scaffold.test.ts
aliasunder: Fixed in b86f7b8 — the test now asserts `composeContent.match(/^\s+image:/gm)` has length 1, so a second service fails it even under a brand-new image name. Kept the retired-image guard and added an MCP_AUTH_TOKEN assertion per the related Sourcery thread.

---

cli/src/docker.ts
aliasunder: Fixed in b86f7b8 — templates.test.ts now asserts `readComposeTemplate("remote")` contains `image: ${GET_TOKEN_IMAGE}`, binding the constant to the image the remote compose actually pulls.

---

cli/src/env.ts
aliasunder: Fixed in b86f7b8 — went one step further than a pinning test: `env.ts` now imports `GET_TOKEN_IMAGE` from `docker.js` and interpolates it, so the literal cannot drift at all, and the env test asserts the generated guidance contains the imported image string (not just "get-token").

---

deploy/remote/README.md
aliasunder: Fixed in b86f7b8 — the docker run section now instructs setting `DEVICE_NAME=vault-cortex` and `CONFLICT_STRATEGY=merge` in `.env` first (both commented in .env.example; DEVICE_NAME to avoid a container-ID device registration, CONFLICT_STRATEGY because the compose default `merge` differs from the CLI default `conflict`). Kept them out of the command as `-e` flags on purpose: with plain docker run, `-e` overrides `--env-file`, which would silently clobber a user's own .env values. SYNC_MODE/LOG_DIR defaults match or are cosmetic, so they stay optional.

---

rootfs/etc/s6-overlay/s6-rc.d/svc-vault-mcp/run
aliasunder: Fixed in b86f7b8 — the run-script comment and the ARCHITECTURE paragraph now state the s6 ordering is slightly weaker than `service_healthy` (no ~30s first-probe delay), that neither ordering ever guaranteed first-sync completion, and that the memory-write shrink guard is the actual fresh-volume defense.

---

rootfs/etc/s6-overlay/scripts/init-setup-user
aliasunder: Fixed in b86f7b8 — `init-setup-user` now records the last-applied `PUID:PGID` in `/home/obsidian/.config/.applied-ids` (config volume, survives recreation) and recursive-chowns whenever the record mismatches — covering the 1500 → 1000 revert the passwd-entry comparison missed. Verified with a three-boot sequence against one volume: unrecorded→1000:1000 (recursive), 1000→1500 (recursive), 1500→1000 (recursive — previously the silent-strand case).

---

deploy/remote/.env.example
aliasunder: Fixed in b86f7b8 — `WINDOWS_MODE: ${WINDOWS_MODE:-false}` added to deploy/remote/docker-compose.yml (mirroring the root Lightsail compose) and CLI templates resynced, so the documented var now reaches the container. The cli/src/env.ts optional-block divergence is pre-existing and lands with the CLI docker-run rework.

---

README.md
aliasunder: Fixed in b86f7b8 — the migration note now splits by setup: local users swap the image line; remote two-service users are told to replace their compose with the new single-service file (with the reason: the standalone sync image no longer receives lockstep releases) and that .env carries over unchanged.

---

cli/src/init.ts
aliasunder: Fixed in b86f7b8 — both comments corrected: the remote flow comment now says "single-service compose", and the compose-up guard comment now describes the real failure mode (init-check-auth fails and s6 stops the container).

---

.env.example
aliasunder: Fixed in b86f7b8 — the VAULT_CORTEX_TAG comment block in .env.example now notes the rename from VAULT_MCP_TAG and that a leftover value under the old name is silently ignored (re-pin under the new name).

---

.github/workflows/deploy.yml
aliasunder: Fixed in b86f7b8 — swapped the step order: remote pushes first, local second, so the local index annotation (the one mirroring server.json) is the most recent on the shared package page under the last-push-wins behavior. Comment documents the ordering contract. Free fix, as suggested — worth keeping even though the GHCR behavior could not be verified before the package exists.

---

aliasunder (commented):
## Phase 1: PR Review

1 finding across 1 file.

64 files reviewed covering: Dockerfile (two-target build), rootfs/ (s6-overlay service definitions), all compose files, all CI workflows, deployment docs, CLI templates/env, tool descriptions, server.json, and package version bumps. Verified all image name references migrated from vault-mcp to vault-cortex, s6 init chain ordering and privilege dropping, Dockerfile stage ordering (local last = default), SHA256 verification of s6-overlay downloads, volume name preservation for seamless migration, and feature surface doc consistency.

**Verdict**: ship-with-minor-fixes

---
*🔍 ship-check · pr-review · claude-opus-4-6[1m]*

.github/workflows/trivy.yml
```diff
@@ -79,33 +124,68 @@ jobs:
...
-          image-ref: vault-mcp:pr-scan
+          image-ref: vault-cortex:pr-scan
           scanners: vuln
           severity: CRITICAL,HIGH
           ignore-unfixed: true
           format: sarif
           output: trivy-results.sarif
           exit-code: "1" # gate: fail the PR on CRITICAL/HIGH findings
 
+      - name: Scan branch image (remote target)
```
aliasunder, line 166: **[D4]** Remote scan skipped when local scan fails — masks remote-only vulnerabilities

The remote scan step has no `if:` condition (defaults to `if: success()`). When the local scan finds a CRITICAL/HIGH vulnerability and fails the step (`exit-code: "1"`), this step is skipped entirely. This means vulnerabilities unique to the remote target’s extra packages (s6-overlay, obsidian-headless, ca-certificates) are masked whenever the shared base image has its own findings.

The remote target is the NEW attack surface this PR introduces to the scan pipeline — not scanning it when the base has issues defeats the purpose of adding it.

Suggested fix: add `if: success() || failure()` (equivalent to `!cancelled()`) so both targets are always scanned independently:

```yaml
      - name: Scan branch image (remote target)
        if: success() || failure()
        uses: aquasecurity/trivy-action@...
```

The `hashFiles` guard on the remote SARIF upload already handles the case where the remote scan didn’t run, so no change needed there.

Would fix (trivial — 1-line `if:` addition).

---
*🔍 ship-check · pr-review · claude-opus-4-6[1m]*

aliasunder (commented):
## Phase 2: Code Quality

2 findings across 2 files. Both are stale references missed during the rename sweep.

---
*🔍 ship-check · code-quality · claude-opus-4-6[1m]*

cli/src/docker.ts
aliasunder, line 14: **[naming]** Stale docstring on line 9 — the `runGetToken` member still says `/** Runs the obsidian-headless-sync get-token flow with inherited stdio. */`. The constant in this file was renamed from `OBSIDIAN_SYNC_IMAGE` to `GET_TOKEN_IMAGE`, and `init.ts` line 63 was updated to say "vault-cortex image's get-token flow" — this docstring was missed.

```suggestion
export const GET_TOKEN_IMAGE = "ghcr.io/aliasunder/vault-cortex:remote"
```

The fix is on line 9 (above the diff hunk), not this line:
```
  /** Runs the vault-cortex get-token flow with inherited stdio. */
```

---
*🔍 ship-check · code-quality · claude-opus-4-6[1m]*

scripts/dev.ts
```diff
@@ -173,9 +175,12 @@ switch (sub) {
...
     run(`scp ${id} ${sshOpts} ${ENV_PATH} ubuntu@${ip}:/opt/vault-cortex/.env`)
     run(
-      `ssh ${id} ${sshOpts} ubuntu@${ip} 'cd /opt/vault-cortex && docker compose pull && docker compose up -d && docker image prune -f'`,
+      `ssh ${id} ${sshOpts} ubuntu@${ip} 'cd /opt/vault-cortex && docker compose pull && docker compose up -d --remove-orphans --wait --wait-timeout 300 && docker image prune -f'`,
     )
-    console.log(`✓ vault-mcp deployed to ${ip}:8000`)
+    // Deliberately no IP in the success line — the instance IP is kept out
+    // of logs (public CI) and is masked above, but not printing it at all is
+    // the stronger guarantee.
+    console.log("✓ vault-cortex deployed (port 8000)")
```
aliasunder, line 183: **[naming]** Stale filename in the default-case usage message at line 189 — `dev.mts` should be `dev.ts`. The file was renamed in a prior commit (`package.json` calls it `scripts/dev.ts`); someone following this usage output would get "file not found." Fix on line 189:

```
      `Usage: tsx scripts/dev.ts <docker:build|docker:push|docker:publish|lightsail:up>`,
```

---
*🔍 ship-check · code-quality · claude-opus-4-6[1m]*

aliasunder (commented):
## Phase 4: Bug Check

1 bug found across 39 production files (28 code/infra/config + 11 docs with factual claims).
Confidence: 1 high, 0 medium, 0 low

Files checked systematically across all 7 dimensions -- full reads of every changed production file, not just the diff. Dimension 1 (description-vs-implementation) received the most scrutiny: tool descriptions in vault-crud-tools.ts verified (cross-section move order, cross-references to vault_get_backlinks / vault_find_orphans / vault_delete_span), config default consistency checked across all 6 docker-compose files + 2 .env.example files + CLI env.ts builders, s6 init chain dependency ordering verified, CI workflow secret scoping / heredoc correctness / concurrency blocks checked.

### **[D1] Tool count mismatch: Dockerfile LABEL + deploy.yml annotation say "25 tools" -- actual count is 26** (pre-existing gap)

`Dockerfile` line 61 and `.github/workflows/deploy.yml` line 179 both say:
> Standalone MCP server for Obsidian vaults -- hybrid search, structured memory, **25 tools** + 3 prompts, OAuth 2.1.

But the actual tool count from the source is **26**:
- `vault-crud-tools.ts`: 9 (read, write, patch, replace, delete-span, list, delete, move, update-properties)
- `search-tools.ts`: 12 (search, search-by-tag, list-tags, recent-notes, search-by-folder, list-property-keys, list-property-values, search-by-property, get-backlinks, get-outgoing-links, find-orphans, list-tasks)
- `memory-tools.ts`: 4 (get-memory, update-memory, list-memory-files, delete-memory)
- `daily-note-tools.ts`: 1 (get-daily-note)

All other descriptions already say 26: `server.json`, `.devin/wiki.json`, `llms-install.md`. The deploy.yml comment on line 176 says "Mirror the server.json / Dockerfile description here" -- but the mirrored text diverges on the count.

Pre-existing on main (both locations already say 25), carried forward unchanged. Fix: `25 tools` to `26 tools` in both `Dockerfile:61` and `.github/workflows/deploy.yml:179`.

*Would fix (trivial, high confidence) -- pre-existing gap*

---
*🔍 ship-check · bug-check · claude-opus-4-6[1m]*

.github/workflows/trivy.yml
aliasunder: Fixed in 8ef3fda — added `if: success() || failure()` to the remote scan step (plus a comment explaining why), so a base-image finding that gates the local scan can no longer mask remote-only CVEs in s6-overlay/obsidian-headless. (Responding to Claude-authored review.)

---

cli/src/docker.ts
aliasunder: Fixed in 8ef3fda — `runGetToken` docstring on line 9 now reads "Runs the vault-cortex get-token flow with inherited stdio." (Responding to Claude-authored review.)

---

scripts/dev.ts
aliasunder: Fixed in 8ef3fda — usage message now says `scripts/dev.ts`, matching the renamed file and package.json. (Responding to Claude-authored review.)

---

PR #312, opened 2026-07-12
fix(search): replace memory recall's absolute rerank floor with adaptive threshold

aliasunder:
## Summary

- Replace the fixed `MEMORY_RECALL_MIN_PROBABILITY = 0.05` sigmoid cutoff with an adaptive 3-parameter floor: `max(sanityFloor, min(maxFloor, bestProbability * relativeRatio))`
- High-confidence queries (best probability near 1.0) keep the 0.05 ceiling (no regression); low-confidence queries follow the best score down to 10% of its value, bounded by a 0.001 sanity floor
- Include the memory file name in the embedding and cross-encoder input (`"Agents > Communication (newest first)\n..."` instead of `"Communication (newest first)\n..."`), giving both models the file name as a relevance signal
- Log `bestProbability` and `effectiveFloor` on every reranked recall for production observability

**Live-verified against production (test deploy):**

| Query | v0.28.1 | This PR |
|---|---|---|
| "how I like agents to communicate with me" | 4 entries (4 of 8 cut) | **11 entries** |
| "opinions on testing" | 48 entries (noisy FTS rescue) | **19 entries** (focused, reranked) |
| "testing" (control) | 33 | 33 |
| nonsense (control) | 0 | 0 |

## Test plan

- [x] Existing 19 memory-recall tests pass unchanged (no regression from adaptive floor)
- [x] New: "keeps a borderline entry when best probability is moderate" (adaptive floor lowers to ~0.027)
- [x] New: "clips the adaptive floor to the max floor for high-confidence queries" (ceiling binds at 0.05)
- [x] Mutation check: setting `RELATIVE_RATIO = 1.0` fails only the borderline test
- [x] Updated rescue log, embedding format, and adaptive-floor log assertions
- [x] Full suite: 1722 tests pass, clean `tsc --noEmit`
- [x] Live verified 4-query matrix against test deploy (results in table above)


aliasunder:
@coderabbit review

src/vault-mcp/search/search-queries.ts
```diff
@@ -445,12 +445,24 @@ const MEMORY_VECTOR_CANDIDATE_LIMIT = 100
...
+ *  Blocks genuinely irrelevant content even when a vague query produces
+ *  universally low cross-encoder scores. Without it, a query whose best
+ *  probability is 0.0003 would set the floor at 0.00003, letting noise
+ *  through. */
+const MEMORY_RECALL_SANITY_FLOOR = 0.001
+
+/** Upper bound on the adaptive relevance floor — preserves the current
+ *  absolute cut for high-confidence queries (best probability near 1.0) so
+ *  this change is a strict non-regression on working queries. */
+const MEMORY_RECALL_ABSOLUTE_FLOOR = 0.05
```
aliasunder: **Naming: `ABSOLUTE_FLOOR` is misleading — it's used as a ceiling, not a floor.**

The formula is `Math.min(MEMORY_RECALL_ABSOLUTE_FLOOR, relativeThreshold)` — the constant caps the adaptive floor from above, making it a ceiling. The JSDoc comment correctly calls it "Upper bound on the adaptive relevance floor," and ARCHITECTURE.md correctly calls it "absolute ceiling." But the name "absolute floor" implies a minimum bound, which is exactly wrong for how it's used.

Consider renaming to `MEMORY_RECALL_FLOOR_CEILING` or `MEMORY_RECALL_ABSOLUTE_CEILING` to match its role as a `Math.min` operand.

---
*🔍 ship-check · pr-review · deepseek/deepseek-v4-pro*

src/vault-mcp/search/search-queries.ts
```diff
@@ -550,17 +564,34 @@ const tryRerankMemoryCandidates = async (
...
+      .filter((candidate) => !candidate.ftsHit)
+      .map((candidate) => probabilityByEntryId.get(candidate.row.id) ?? 0)
+    const bestProbability =
+      nonFtsHitProbabilities.length > 0
+        ? Math.max(...nonFtsHitProbabilities)
+        : MEMORY_RECALL_ABSOLUTE_FLOOR
+    const relativeThreshold = bestProbability * MEMORY_RECALL_RELATIVE_RATIO
+    const effectiveFloor = Math.max(
+      MEMORY_RECALL_SANITY_FLOOR,
+      Math.min(MEMORY_RECALL_ABSOLUTE_FLOOR, relativeThreshold),
```
aliasunder: **Usage reinforces the naming confusion:** `Math.min(MEMORY_RECALL_ABSOLUTE_FLOOR, ...)` reads as "take the minimum of a floor" — a floor is already a minimum, so this reads as a double negative. With the renamed constant from the finding above, this would read as `Math.min(FLOOR_CEILING, relativeThreshold)` which is clearer about the clamping intent.

---
*🔍 ship-check · pr-review · deepseek/deepseek-v4-pro*

aliasunder (commented):
## PR Review: #312 — Adaptive rerank floor

### Summary
This is a clean, well-scoped change. The adaptive floor formula is correct, the test coverage is solid, and the architecture docs are accurate. The one substantive finding is a naming issue.

### Findings

**1. Naming — MEMORY_RECALL_ABSOLUTE_FLOOR is misleading** (would-fix)
The constant is used as a ceiling in the adaptive formula (Math.min(ABSOLUTE_FLOOR, relativeThreshold)), but "absolute floor" suggests a minimum bound. The JSDoc and ARCHITECTURE.md both correctly describe it as an "upper bound" / "ceiling." The name should match the role — MEMORY_RECALL_FLOOR_CEILING or MEMORY_RECALL_ABSOLUTE_CEILING would be accurate.

**2. No stale references, no feature surface doc gap**
MEMORY_RECALL_MIN_PROBABILITY is fully removed. Tool counts unchanged — server.json and README.md don't need updates.

**3. Formula correctness: ✓**
All traces (best=1.0, 0.5, 0.3, 0.01, 0, negative) produce correct results. The sanity floor (0.001) correctly guards against degenerate best probabilities. The bestProbability fallback to ABSOLUTE_FLOOR when no non-ftsHit candidates exist is sound.

**4. Test quality: ✓**
The new test is well-constructed — inline mock reranker, exact assertions, and the test name accurately describes the adaptive floor behavior. The existing log assertions were correctly updated with the new diagnostics fields.

### Verdict: ship-with-minor-fixes
The naming issue is the only actionable finding. The logic is correct, the tests pass, and the docs are accurate. Not a blocker.

---
*🔍 ship-check · pr-review · deepseek/deepseek-v4-pro*

ARCHITECTURE.md
```diff
@@ -246,9 +246,11 @@ content-hash identity rather than position, so a newest-first top-insert
 re-embeds exactly one entry.
 
 The query unions every lexical match with the vector top-100, fused by RRF,
-then cuts on an absolute cross-encoder relevance floor (sigmoid ≥ 0.05; lexical
-hits always survive) — evolution-arc origins with drifted vocabulary are kept
-while vague queries stay bounded. Degrades to a distance-margin cut without the
+then cuts on an adaptive cross-encoder relevance floor (clamped between a
```
aliasunder: Nit: The docs correctly call it "0.05 absolute ceiling" — this is more accurate than the code's naming of "ABSOLUTE_FLOOR." The docs and code should agree on terminology. If the constant is renamed (see finding above), the docs are already correct and need no change.

---
*🔍 ship-check · pr-review · deepseek/deepseek-v4-pro*

aliasunder (commented):
## Code quality review — search-queries.ts

One finding worth flagging, everything else is clean.

### Finding: `MEMORY_RECALL_ABSOLUTE_FLOOR` serves double duty

At line 576 the constant is used as a **default `bestProbability`** when every candidate is a lexical hit (no nonFtsHit data points to compute from). At line 580 it's used as the **ceiling in the clamp** — its primary role per the doc comment.

These are semantically different purposes:
- "Upper bound on the adaptive floor" (ceiling)
- "Reasonable best-probability value when we have no vector-only candidates to judge" (default)

Reusing the same constant couples them: changing the ceiling to 0.08 would silently raise the default bestProbability to 0.08, making `relativeThreshold = 0.008` instead of `0.005`. Whether that's correct depends on whether the ceiling and the default should move together.

A separate constant (e.g. `MEMORY_RECALL_DEFAULT_PROBABILITY = 0.05` or even just `1.0` — "every candidate is already a lexical hit, so the floor shouldn't matter") would keep the two concerns independent.

(Phase 1 already flagged the floor-as-ceiling naming mismatch; this is the companion finding — same constant, second role.)

### What looks good
- **Comments**: new constants have clear "why" doc comments; the adaptive-floor comment at line 568 explains strategy well
- **Simplicity**: the clamp chain (`Math.max(sanity, Math.min(ceiling, relative))`) is clean and readable
- **Structure**: `diagnostics` struct replacing a bare boolean is a clean refactor — easy to extend
- **No new dependencies**: everything stays within the search module
- **Immutability**: no `let`, no mutation, clean functional style

**Verdict**: merge-ready; the finding is a naming/coupling nit, not a bug.

---

*🔍 ship-check · code-quality · deepseek/deepseek-v4-pro*

src/vault-mcp/search/search-queries.ts
```diff
@@ -550,17 +564,34 @@ const tryRerankMemoryCandidates = async (
...
+
+    // Adaptive floor: follows the best score down when overall confidence
+    // is moderate, bounded between the sanity floor and the absolute floor.
+    const nonFtsHitProbabilities = candidates
+      .filter((candidate) => !candidate.ftsHit)
+      .map((candidate) => probabilityByEntryId.get(candidate.row.id) ?? 0)
+    const bestProbability =
+      nonFtsHitProbabilities.length > 0
+        ? Math.max(...nonFtsHitProbabilities)
+        : MEMORY_RECALL_ABSOLUTE_FLOOR
```
aliasunder: **Naming/coupling**: `MEMORY_RECALL_ABSOLUTE_FLOOR` serves double duty — it's used as the ceiling in the clamp (line 580, its primary role per the doc comment) and also as a default `bestProbability` when every candidate is a lexical hit (here, line 576).

Reusing the same constant couples two independent decisions: changing the ceiling would silently change the default probability. A separate constant (e.g. `MEMORY_RECALL_DEFAULT_PROBABILITY`) would keep them decoupled. Not a bug, but worth noting alongside the Phase 1 finding that this constant's name says "floor" while it acts as a ceiling in `Math.min`.

---
*🔍 ship-check · code-quality · deepseek/deepseek-v4-pro*

src/vault-mcp/search/__tests__/memory-recall.test.ts
```diff
@@ -451,11 +451,62 @@ describe("memoryRecall", () => {
           matched: 1,
           returned: 1,
           anyTermRescue: true,
+          bestProbability: sigmoid(-8),
+          effectiveFloor: 0.001,
         },
       ],
     ])
   })
```
aliasunder, line 458: ## Coverage gap: absolute-ceiling branch of adaptive floor not tested

The adaptive floor expression `max(sanityFloor, min(absFloor, relativeThreshold))` has three branches, but the test suite covers only two:

| Branch | Test |
|---|---|
| Sanity floor wins (rel < 0.001) | Rescue test (all logits -8) ✓ |
| Relative ratio wins (0.001 < rel < 0.05) | New test (best logit -1) ✓ |
| **Absolute ceiling wins (rel > 0.05)** | **None** ✗ |

This branch activates when `bestProbability > 0.5` — i.e., a high-confidence query where `bestProbability * RELATIVE_RATIO` exceeds 0.05. The `min(ABSOLUTE_FLOOR, ...)` guard is the regression anchor for working queries (per the constant's doc comment), but if that guard is accidentally removed or `ABSOLUTE_FLOOR` is raised, this branch silently absorbs the error.

**Suggested fix**: Add a test with a reranker that produces a confident best logit (e.g., 3 → sigmoid ≈ 0.95, rel ≈ 0.095 > 0.05) and assert that the effectiveFloor clips to 0.05, not 0.095. An entry with sigmoid(-3) ≈ 0.047 would still be cut (below 0.05), confirming the ceiling binds.

**Mutation check**: If `ABSOLUTE_FLOOR` were raised to 0.10, only a test in this branch would fail — the two existing branches would still pass since their relativeThreshold values (0.00003 and 0.027) are both below 0.10.

---
*🔍 ship-check · test-audit · deepseek/deepseek-v4-pro*

aliasunder (commented):
## Test-audit review (Phase 3)

### Summary

The new test is solid — passes both bars of the two-bar rule, uses exact `toEqual` assertions, and correctly exercises the happy path of the adaptive floor. The rescue test update for new log fields is also correct.

### Findings

**1 coverage gap** found (posted inline). The non-rescue log fields gap is noted below but not flagged — it's a soft observability gap, not a correctness risk.

### Detailed evaluation

**New test** (`keeps a borderline entry that the absolute floor would cut…`):
- ✓ Two-bar: fails if adaptive floor logic breaks (only 1 of 2 entries survives); passes only for intended reason (sigmoid(-3) ≈ 0.047 < 0.05 old floor, > 0.027 new adaptive floor)
- ✓ Assertions: `toEqual` on full date array, `toBe(true)` on reranked
- ✓ Hygiene: `const` per test, explicit callback names, test name matches assertion
- ✓ Mutation: setting `RELATIVE_RATIO = 1.0` causes the borderline entry to be cut → test fails correctly

**Rescue test update**: `sigmoid(-8)` import avoids hardcoded magic numbers; exact `toEqual` on the full log object includes both new fields. The effectiveFloor assertion (0.001) covers the sanity-floor branch.

**Non-rescue log path** (`logHybridResult` at search-queries.ts:833): The normal hybrid path now emits `bestProbability` and `effectiveFloor`, but no test asserts these fields in the non-rescue code path. This is a soft gap — entry behavior is still tested, but observability contract isn't.

---
*🔍 ship-check · test-audit · deepseek/deepseek-v4-pro*

src/vault-mcp/search/__tests__/memory-recall.test.ts
```diff
@@ -451,11 +451,62 @@ describe("memoryRecall", () => {
           matched: 1,
           returned: 1,
           anyTermRescue: true,
+          bestProbability: sigmoid(-8),
+          effectiveFloor: 0.001,
```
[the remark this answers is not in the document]
aliasunder: Fixed in 66fb1be — added logger assertion to the borderline test verifying `bestProbability: sigmoid(-1)` and `effectiveFloor: sigmoid(-1) * 0.1`, so the test fails if the computation degenerates to the sanity floor. Also added a separate ceiling-branch test per the test-audit finding.

---
*🔍 ship-check · pr-monitor · Claude Opus 4.6 (1M context)*

src/vault-mcp/search/search-queries.ts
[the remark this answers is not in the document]
aliasunder: This is intentional — when all candidates are ftsHit, the floor is never consulted (ftsHit bypasses it unconditionally at the filter). The bestProbability value in this branch is only used for logging diagnostics, and MEMORY_RECALL_MAX_FLOOR (now renamed from ABSOLUTE_FLOOR) is a reasonable sentinel meaning "standard operating level" in the log. Setting it to 0 would produce a misleading log line (bestProbability: 0, effectiveFloor: 0.001) that looks like a degenerate result when the actual result is fine.

---
*🔍 ship-check · pr-monitor · Claude Opus 4.6 (1M context)*

src/vault-mcp/search/__tests__/memory-recall.test.ts
```diff
@@ -451,11 +451,62 @@ describe("memoryRecall", () => {
...
+      logger,
+    )
+    expect(reranked).toBe(true)
+    // Both communication entries survive: "structured" at sigmoid(-3) ≈ 0.047
+    // would be cut by the old absolute floor (0.05) but the adaptive floor
+    // lowers to ~0.027 when the best probability is only ~0.27.
+    expect(entries.map((entry) => entry.date)).toEqual([
+      "2026-06-15",
+      "2026-07-01",
+    ])
```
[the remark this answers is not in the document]
aliasunder: This is intentional — `buildMemoryRecallResult` sorts the final output chronologically via `compareMemoryEntriesChronologically` (search-queries.ts:576-582), not by relevance. Relevance ordering is only used for truncation (dropping the least-relevant entries when `maxResults` is exceeded). The date order `["2026-06-15", "2026-07-01"]` is the correct chronological output.

---
*🔍 ship-check · pr-monitor · Claude Opus 4.6 (1M context)*

src/vault-mcp/search/search-queries.ts
aliasunder: Fixed in 66fb1be — renamed `MEMORY_RECALL_ABSOLUTE_FLOOR` to `MEMORY_RECALL_MAX_FLOOR` across all 4 occurrences. "Max floor" reads naturally: `Math.min(MAX_FLOOR, relativeThreshold)` — take the smaller of the max floor and the relative threshold. Doc comment updated to "Maximum value the adaptive floor can take." (Responding to Claude-authored review.)

---
*🔍 ship-check · pr-monitor · Claude Opus 4.6 (1M context)*

src/vault-mcp/search/search-queries.ts
aliasunder: Resolved by the rename — `Math.min(MEMORY_RECALL_MAX_FLOOR, relativeThreshold)` reads as "take the smaller of the max floor and the relative threshold," which is exactly what it does. (Responding to Claude-authored review.)

---
*🔍 ship-check · pr-monitor · Claude Opus 4.6 (1M context)*

ARCHITECTURE.md
aliasunder: ARCHITECTURE.md already uses "absolute ceiling" which is consistent with the renamed `MEMORY_RECALL_MAX_FLOOR` — no doc change needed. (Responding to Claude-authored review.)

---
*🔍 ship-check · pr-monitor · Claude Opus 4.6 (1M context)*

src/vault-mcp/search/search-queries.ts
aliasunder: The coupling finding is accurate structurally, but in practice the two uses are not independent: when all candidates are ftsHit, the floor is never consulted (ftsHit bypasses it unconditionally). The bestProbability value in that branch is purely a logging sentinel. Introducing a separate `MEMORY_RECALL_DEFAULT_PROBABILITY` constant for a value that only appears in diagnostic logs and never affects kept/rejected decisions adds a declaration to maintain with no behavioral benefit. If the ceiling ever needs to decouple from the logging default, that is the time to split. (Responding to Claude-authored review.)

---
*🔍 ship-check · pr-monitor · Claude Opus 4.6 (1M context)*

src/vault-mcp/search/__tests__/memory-recall.test.ts
aliasunder: Fixed in 66fb1be — added "clips the adaptive floor to the max floor for high-confidence queries" test: reranker returns logit 3 (sigmoid 0.95) for the best entry and logit -3 (sigmoid 0.047) for a borderline entry. Asserts effectiveFloor clips to 0.05 (the ceiling binds), and the borderline entry at 0.047 < 0.05 is correctly cut. All three branches of the adaptive floor formula are now covered. (Responding to Claude-authored review.)

---
*🔍 ship-check · pr-monitor · Claude Opus 4.6 (1M context)*

aliasunder:
## Bot review body findings — addressed

**Sourcery body (2 overall comments):**
1. "nonFtsHitProbabilities defaults to ABSOLUTE_FLOOR when empty" — intentional; the value is purely a logging sentinel (ftsHit candidates bypass the floor unconditionally). Replied inline on the matching thread.
2. "missing probabilities coerced to 0 can lower bestProbability" — the `?? 0` coercion handles the logits/candidates length mismatch case documented as "cannot normally happen." A probability-of-0 entry is correctly treated as maximally irrelevant. Filtering them out instead would silently ignore the mismatch; coercing to 0 makes them visible in the log while sorting them last.

**CodeRabbit nitpick (lines 461-509):**
Assert `bestProbability`/`effectiveFloor` in the borderline test — fixed in 66fb1be with exact log assertions on both the borderline test and a new ceiling-branch test.

**Ship-check reviews (pr-review, code-quality, test-audit):**
All findings overlap with inline threads and are addressed there. The naming rename (`ABSOLUTE_FLOOR` → `MAX_FLOOR`) and the ceiling-branch coverage gap are both fixed.

---
*🔍 ship-check · pr-monitor · Claude Opus 4.6 (1M context)*

PR #372, opened 2026-07-24
feat: surface leading content in outline and warn on displacing prepends

aliasunder:
## Problem

A file-level `vault_patch_note` prepend (no `heading` target) inserts at body line 0. When the inserted content starts with a heading and the note already has content above its first heading, that content silently becomes the new section's body.

Nothing surfaced it: `outline` mode reported only `leading_callout`, never plain leading text, so the displaced region was invisible to every structured read. In the incident that prompted this, an activity log's intro block stayed displaced for 2.5 weeks before anyone noticed.

`insert_before` was the correct operation and already existed — the API just made the wrong one easy and the damage silent.

## The precise hazard

Section spans run to the next heading of **same or higher** level, so a prepend only restructures content when all of:

1. `operation === "prepend"`, and
2. no `heading` target, and
3. the region above the first heading has a non-blank line, and
4. the inserted content's first non-blank line is itself a heading.

There is deliberately no "note has ≥1 heading" clause — prepending a heading onto a heading-less prose note pulls the whole body in, which is the same displacement. Everything else is harmless, including the documented leading-callout workflow, which the negative tests pin.

## Changes

**`leading_content` in outline** — the body text above the first heading, with any leading callout's own lines excluded so the two fields never repeat the same bytes. Together they cover the whole region. Omitted when absent, mirroring `leading_callout`.

The callout is subtracted by index filter rather than span arithmetic, which keeps the callout-after-a-leading-H1 case safe: that span falls outside the region entirely, so no index matches and a negative slice is structurally impossible.

**Write-time advisory** — the confirmation gains a sentence when a prepend displaces leading content. The write still succeeds; this is advisory, not a rejection. Two variants, because condition 3 above admits a case with no `insert_before` target:

```
Applied prepend to Logs/Activity.md → file body. The inserted heading now
contains 412 bytes of content that already sat above the note's first heading.
To place a section above that content instead, use operation "insert_before"
with heading "Week 27" (H2).

Applied prepend to Notes/Idea.md → file body. The inserted heading now contains
the note's entire body (287 bytes) — the note had no headings of its own. To
place a section below that content instead, use operation "append".
```

The heading is named by its **bare** text, which is what the `heading` param matches — the `##`-prefixed display form would send a caller straight into `heading not found`. Exact-string tests lock both variants.

Composed in the handler, not the data layer: the advisory names `insert_before`/`append`, which are MCP surfaces, and data-layer code must not name those or prescribe caller-level remediation.

**Tool descriptions** — the `prepend:` operation line routes to `insert_before`; a `Limitation:` block names the displacement. The advisory is deliberately *not* under `Errors:`, since callers read that section as failure modes and would retry a successful write. A test enforces that placement.

## Two bugs fixed along the way

**Trailing comment blocks leaked into the region.** With no headings, the region ran to EOF, so an empty Kanban board — no lanes yet, just its `%% kanban:settings %%` block — reported the settings JSON as body content and would have fired a spurious advisory on the first lane added. The region now stops at `findTrailingCommentBlockStart`, mirroring how `parseHeadings` bounds a final section.

**CRLF headings went undetected.** `HEADING_REGEX` ends in `(.+)$` and `.` excludes CR, so `"## New\r"` matched nothing. `leadingHeadingOfContent` now strips a trailing CR before parsing. This was extracted from the existing duplicate-heading guard, which had the identical blind spot — a CRLF-authored heading bypassed it entirely.

## Verification

- 2112 tests pass; `npm run build`, `lint` (0 errors), `prettier:check` all green.
- **Mutation-checked**, each confirmed to fail for its own reason: region running to EOF (kanban tests fail), callout span ending at the trimmed body (blank-quote tests fail), advisory firing unconditionally (all 8 negative cases fail), and the file-level write being skipped entirely (**every** negative advisory test fails — so none of them can pass via silent no-op).
- One existing test changed: `readNoteOutline` "returns an empty headings array for a note with no headings" now also carries `leading_content`, and was renamed to match what it asserts. Two neighbouring tests were upgraded from decomposed assertions to whole-object equality so a spurious `leading_content` fails them too.

## Compatibility

Both response changes are additive — `leading_content` is a new optional key, and the confirmation is untyped text gaining a trailing sentence. `patchNote` and `readNoteOutline` each had exactly one production caller, so the widened internal return types are fully contained.

## Known gaps, not addressed here

`HEADING_REGEX` doesn't recognize tab-separated (`##\tTitle`), 3-space-indented, or setext headings, all of which Obsidian accepts. Pre-existing, but this PR widens their blast radius slightly: such a heading now also escapes the advisory and lands in `leading_content` as body text. Worth its own change.

https://claude.ai/code/session_01HBUcX98HzBrat3wJ1kMnkD

## Summary by CodeRabbit

* **New Features**
  * Note outlines now include leading body content separately from leading callouts.
  * Note patching provides clearer guidance when prepending content displaces existing introductory text.
  * Heading-targeted and file-level patch operations now offer more consistent behavior and feedback.

* **Documentation**
  * Updated guidance for note reading, patching, replacing, and deleting content to clarify supported operations and response formats.


aliasunder:
@coderabbit review

PR #382, opened 2026-07-29
fix: make upsertNote and removeNote atomic; state busy_timeout explicitly

aliasunder:
## What

Two hardening changes to `search-index.ts` (task: SQLite index robustness), plus one same-class fix found during planning:

1. **`upsertNote` is now transactional.** Its full statement sequence (FTS delete → notes upsert → FTS insert → tasks delete/insert → memory entries → links delete/insert/re-resolve) runs inside `db.transaction(() => { ... })()`. A mid-sequence throw previously left notes/FTS/tasks/links mutually inconsistent until the next full reindex; it now rolls back cleanly. Pure computation (`metadataText`, `taskFolder`, `extractedTasks`, `memoryFile`) is hoisted above the wrap so the write lock isn't held during parsing. `links.extractAll` deliberately stays inside: it only runs when `!skipLinks`, and rebuild Pass 1 calls every note with `skipLinks: true` — hoisting would add wasted per-note extraction to every rebuild.
2. **`removeNote` gets the same wrap** — it ran 4–6 bare deletes, so a mid-sequence failure left a note half-removed. Its `removeMemoryEntriesForFile` call (itself non-transactional) now also participates in the atomic scope.
3. **`busy_timeout` stated explicitly.** Planning verified the original premise was wrong: better-sqlite3 already applies `sqlite3_busy_timeout(handle, 5000)` at open by default (`lib/database.js:34`, native `database.cpp:172`), so no behavior changes. The explicit pragma + comment exist so the retry contract survives refactors (e.g. a future `timeout: 0` open option) and isn't re-flagged by audits. Per that finding, no dedicated test — a worker-thread contention harness to verify a library default was judged disproportionate.

## Design notes

- **Default (deferred) transaction variant**, consistent with the four existing `db.transaction()` uses. Background: `busy_timeout`'s retry handler only protects plain write-lock acquisition — it deliberately does not retry a deferred transaction that *reads before its first write*. There, the read pins a WAL snapshot; if another connection commits meanwhile, the snapshot is stale and the later write-lock upgrade fails instantly with `SQLITE_BUSY_SNAPSHOT`, ignoring the timeout (retrying can't refresh a mid-transaction snapshot — only rollback-and-restart can). Both new wraps open with a DELETE — a write — so the write lock is contested at statement 1, where the busy handler does apply and waits the full 5000 ms. The write-first shape makes the stale-snapshot window structurally impossible, giving the deferred default the same safety as `.immediate()` here. The caveat is recorded in a comment on the `busy_timeout` pragma for future wraps (a wrap that must read before its first write should use `.immediate()`).
- **Nesting is safe**: `rebuildFromVault` (own tx) → `upsertNote` → `upsertMemoryEntries` (own tx) becomes savepoint nesting, which better-sqlite3 handles natively; no manual `BEGIN`/`COMMIT` exists anywhere in `src/`.
- The `if (skipLinks) return` now exits the transaction callback (return = commit, throw = rollback); a comment on the wrap notes that nothing may run between the wrap's `)()` and the function's end.
- The OAuth DB relies on the same 5000 ms driver default — same non-gap, consciously left untouched.

## Tests

Three atomicity tests via a statement-poisoning seam: a prototype-level `prepare` spy (installed before `createSearchIndex`, since all write statements are prepared at factory scope) patches one statement's `.run` to throw while armed, so seeding runs clean and the failure fires mid-sequence on demand.

- **Update-over-existing rolls back**: version B's upsert throws after its notes/FTS writes already ran in-transaction; version A asserted fully intact (exact whole-value assertions: `recentNotes`, `fullTextSearch`, `listTasks`, `getOutgoingLinks`).
- **Fresh insert leaves no trace**, with a control note guarding against `fullTextSearch`'s catch-and-return-`[]` making absence assertions vacuous.
- **removeNote keeps the note fully present**: poisons the last unconditional delete, so the FTS/notes/links deletes had all run when it fired — full presence proves rollback.

**Mutation-tested**: each wrap was temporarily replaced with a bare IIFE; the matching tests failed on partial-state assertions (version B persisted / note vanished despite the throw), not on the throw assertion, then the wraps were restored.

## Verification

- `npm test`: 2159 passed, 3 consecutive full-suite runs (one unidentified single-test failure on the very first run did not reproduce across three reruns — suspected pre-existing flake; watching CI)
- `npm run lint`: 0 errors, 336 warnings (identical count to main — none introduced)
- `npm run build`: clean

No user-facing surface changes: no env vars, no tool descriptions, no docs impact.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search index reliability by making note updates and removals atomic.
  * Prevented partially completed changes from leaving behind inconsistent search results, tasks, links, vectors, or memory entries.
  * Failed operations now roll back cleanly, preserving the previously saved state or leaving no residual data for new notes.
  * Improved database handling during temporary contention.

### DIFF
--- a/src/vault-mcp/search/__tests__/search-index.test.ts
+++ b/src/vault-mcp/search/__tests__/search-index.test.ts
@@ -528,7 +528,7 @@ describe("upsertNote", () => {
 
 describe("upsertNote atomicity", () => {
   it("rolls back to the prior version when a statement fails mid-upsert", () => {
-    const poison = installStatementPoison("INSERT INTO tasks")
+    const taskInsertPoison = installStatementPoison("INSERT INTO tasks")
     const atomicIndex = createSearchIndex(":memory:")
     atomicIndex.upsertNote(
       {
@@ -543,7 +543,7 @@ describe("upsertNote atomicity", () => {
     // notes upsert, FTS insert, and tasks delete have already run inside the
     // transaction — intact version-A state below proves genuine rollback,
     // not a no-op.
-    poison.arm()
+    taskInsertPoison.arm()
     expect(() =>
       atomicIndex.upsertNote(
         {
@@ -553,8 +553,8 @@ describe("upsertNote atomicity", () => {
         },
         logger,
       ),
-    ).toThrow(poison.message)
-    poison.disarm()
+    ).toThrow(taskInsertPoison.message)
+    taskInsertPoison.disarm()
 
     expect(atomicIndex.recentNotes({}, logger)).toEqual([versionAMetadata()])
     const versionAHits = atomicIndex.fullTextSearch(
@@ -577,7 +577,9 @@ describe("upsertNote atomicity", () => {
   })
 
   it("rolls back the link phase when a link statement fails mid-upsert", () => {
-    const poison = installStatementPoison("INSERT OR IGNORE INTO links")
+    const linkInsertPoison = installStatementPoison(
+      "INSERT OR IGNORE INTO links",
+    )
     const atomicIndex = createSearchIndex(":memory:")
     atomicIndex.upsertNote(
       {
@@ -593,7 +595,7 @@ describe("upsertNote atomicity", () => {
     // version-A link below proves link-phase rollback specifically, which
     // the tasks-phase poison above never exercises (it throws before the
     // links table is touched).
-    poison.arm()
+    linkInsertPoison.arm()
     expect(() =>
       atomicIndex.upsertNote(
         {
@@ -603,8 +605,8 @@ describe("upsertNote atomicity", () => {
         },
         logger,
       ),
-    ).toThrow(poison.message)
-    poison.disarm()
+    ).toThrow(linkInsertPoison.message)
+    linkInsertPoison.disarm()
 
     expect(atomicIndex.recentNotes({}, logger)).toEqual([versionAMetadata()])
     const versionAHits = atomicIndex.fullTextSearch(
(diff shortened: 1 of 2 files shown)


aliasunder:
@coderabbit review

aliasunder:
Addressed all three CodeRabbit review-body nitpicks in 41eda07:

1. **Success log inside the transaction** (search-index.ts:1111): Valid — link extraction could throw and roll everything back after `indexed note` was already emitted. Moved the debug log after the transaction's `)()` so a rolled-back write can't leave a success line; it still fires on both the `skipLinks` and full paths, and the transaction comment's post-`)()` constraint was reworded to scope it to link-phase work.

2. **Inline prepare in `rebuildFromVault` Pass 2** (search-index.ts:1496): Valid — duplicated the factory-scope `selectAllNotePathsStmt` hoisted in the previous commit. Pass 2 now reuses the hoisted statement.

3. **Link rollback not covered by the tasks-phase poison** (search-index.test.ts:574): Valid — `INSERT INTO tasks` fires before `deleteLinksStmt`, so the links table was never mutated in the poisoned run. Added a third atomicity test poisoning `INSERT OR IGNORE INTO links`; by then `deleteLinksStmt` has already cleared version A's links, so the intact version-A link proves link-phase rollback specifically. Mutation-tested: with only the link delete hoisted outside the transaction, the test fails on the link assertion (`expected [] to equal [Alpha Target]`), not an earlier one.

2188 tests, lint and build clean.

---

PR #468, opened 2026-08-20
fix(security): gate Forwarded/X-Forwarded-For client-IP trust behind explicit proxy config

aliasunder:
## What does this PR do?

Fixes GHSA-wm5v-9236-597m: the OAuth route rate limiter derived its bucket key from the client-supplied `Forwarded` header with no trusted-proxy validation, so an attacker could mint a fresh rate-limit bucket per request. The fix closes the sibling X-Forwarded-For channel as well.

**Fix:** client-IP trust is now deployment-explicit. `TRUST_PROXY_HOPS` (default `0` — direct exposure) feeds Express `trust proxy`; `TRUST_FORWARDED_HEADER` (default `false`) gates the RFC 7239 `Forwarded` header for proxies that set it themselves (API Gateway). When trusted, `extractClientIp` reads the **last** `for=` element — correct whether the edge proxy replaces or appends (API Gateway appends; with first-element reads a client-supplied prefix would still win, even on the documented AWS path). The AWS reference compose sets both switches; proxied self-hosters get `TRUST_PROXY_HOPS=1` via updated deploy docs.

## Type of change

- [ ] New MCP tool
- [x] Bug fix
- [ ] Refactoring
- [ ] Documentation
- [ ] CI / workflow change
- [ ] Infrastructure (SST, Docker)
- [ ] Other: 

## Checklist

- [x] `npm test` passes (2881 tests)
- [x] `npm run lint` passes
- [x] `npm run prettier:check` passes
- [x] `npm run build` succeeds
- [x] New MCP tools follow the naming and description conventions in AGENTS.md — N/A (no new tools)
- [x] README or ARCHITECTURE.md updated (README env-var table, deploy guides, DEPLOY.md)

BREAKING CHANGE: self-hosted deployments behind a reverse proxy or tunnel (Caddy, nginx, Cloudflare Tunnel) must set `TRUST_PROXY_HOPS=1` — without it, all clients share the proxy's OAuth rate-limit budget. The AWS reference deployment and direct-exposure setups need no action.


docker-compose.yml
```diff
@@ -41,6 +41,12 @@ services:
       INDEX_DB_PATH: /data/index.db
       MCP_AUTH_TOKEN: ${MCP_AUTH_TOKEN:?MCP_AUTH_TOKEN required for in-process auth}
       PUBLIC_URL: ${PUBLIC_URL:?PUBLIC_URL required for OAuth metadata}
+      # API Gateway is the immediate upstream: its RFC 7239 Forwarded claim is
+      # the client identity for rate limiting. With ORIGIN_URL (a tunnel/proxy
+      # BETWEEN the gateway and this container) set TRUST_FORWARDED_HEADER=false
+      # and TRUST_PROXY_HOPS=2 — the X-Forwarded-For chain carries the client.
+      TRUST_FORWARDED_HEADER: ${TRUST_FORWARDED_HEADER:-true}
```
aliasunder, line 51: **[D6]** The root `docker-compose.yml` correctly passes these vars for the Lightsail reference deployment. However, the corresponding deploy workflows (`.github/workflows/deploy.yml` and `test_deploy.yml`) don't write `TRUST_PROXY_HOPS` or `TRUST_FORWARDED_HEADER` to the Lightsail `.env` — so the compose `${VAR:-default}` interpolation is the only source. This works for the reference deployment (defaults are correct), but the `ORIGIN_URL` case documented in this PR's `.env.example` (requiring `false`/`2`) can't be deployed via the workflow. See the review body for full details and suggested fix.

Would fix.

---
*🔍 ship-check · pr-review · claude-opus-4-6[1m]*

aliasunder (commented):
## Phase 1: PR Review

5 findings across 4 files. The core auth logic and security model are sound — the last-for= approach is correct, defaults are safe, and the integration tests directly verify the GHSA regression. Findings are all doc/deploy-surface gaps.

### Inline findings

See inline comment on `docker-compose.yml`.

### Beyond-diff findings

**[D6] `ARCHITECTURE.md:586` — stale rate limiting description**

The rate limiting paragraph says the key generator "extracts the real client IP from API Gateway's `Forwarded` header." This is now conditional on `TRUST_FORWARDED_HEADER`. Suggested update:

```
A custom key generator extracts the real client IP via `extractClientIp` —
when `TRUST_FORWARDED_HEADER` is set (API Gateway deployments), the last
`for=` element of the RFC 7239 Forwarded header is the bucket key; otherwise
`req.ip`, governed by Express `trust proxy` hop count (`TRUST_PROXY_HOPS`),
is the key
```
Would fix (trivial).

**[D6] `ARCHITECTURE.md:704` — stale security table entry**

The security table says "`trust proxy` = 1 (Express) | Trusts exactly one proxy hop (API Gateway)" — this is now `TRUST_PROXY_HOPS` (default 0 for direct exposure). Suggested replacement:

```
| `trust proxy` = TRUST_PROXY_HOPS | Default 0 (direct exposure — req.ip is the socket peer, injected X-Forwarded-For ignored); set to the number of trusted proxy hops for reverse-proxy deployments |
```
Would fix (trivial).

**[D6] `.github/workflows/deploy.yml` + `.github/workflows/test_deploy.yml` — missing `TRUST_PROXY_HOPS` / `TRUST_FORWARDED_HEADER` passthrough**

Neither deploy workflow passes these new env vars through to the Lightsail `.env`. Per env-var checklist step 7 in AGENTS.md: "A var the workflows never write can never reach the instance." The compose defaults handle the reference deployment correctly, but the `ORIGIN_URL` case documented in this PR's own `.env.example` — requiring `TRUST_FORWARDED_HEADER=false` and `TRUST_PROXY_HOPS=2` — can't be deployed through the workflow. An operator who sets these manually on the instance would have them overwritten on the next deploy.

Both workflows need:
1. Env vars in the step's `env:` block: `TRUST_PROXY_HOPS: ${{ vars.TRUST_PROXY_HOPS }}` and `TRUST_FORWARDED_HEADER: ${{ vars.TRUST_FORWARDED_HEADER }}`
2. Conditional-write lines: `if [ -n "$TRUST_PROXY_HOPS" ]; then echo "TRUST_PROXY_HOPS=$TRUST_PROXY_HOPS" >> "$RUNNER_TEMP/lightsail.env"; fi` (and same for `TRUST_FORWARDED_HEADER`)

Would fix.

**[D6] `docker-compose.local.yml` — missing `TRUST_PROXY_HOPS` / `TRUST_FORWARDED_HEADER`**

This file lists every env var the server reads. The server defaults (0/false) are correct for local dev so no runtime bug, but the omission breaks the established pattern. Add after `DISABLED_TOOLS`:
```yaml
      TRUST_PROXY_HOPS: ${TRUST_PROXY_HOPS:-0}
      TRUST_FORWARDED_HEADER: ${TRUST_FORWARDED_HEADER:-false}
```
Would fix (trivial).

**[D1] `src/vault-mcp/mcp-core/__tests__/mcp-router.test.ts:169` — pre-existing: harness hardcodes `trust proxy: true`**

The test harness hardcodes `app.set("trust proxy", true)` instead of using `config.trustProxyHops`. The new "ignores Forwarded when not trusted" test (line 674) asserts `clientIp: FORWARDED_IP` ("192.0.2.10" from XFF in `baseHeaders`), which passes because Express trusts XFF. In production with default config (`trustProxyHops: 0`), `req.ip` would be the socket peer (e.g. `127.0.0.1`). The test verifies the right security property (Forwarded header is ignored) but with a wrong intermediate value (req.ip from XFF rather than socket peer).

Flagged: pre-existing gap. Changing the harness would cascade to existing tests.

**Verdict**: ship-with-minor-fixes

---
*🔍 ship-check · pr-review · claude-opus-4-6[1m]*

src/auth.ts
```diff
@@ -30,19 +30,39 @@ export const parseBearer = (header: string | undefined): string | null => {
...
+ * prepend its own elements, so the first `for=` is attacker-controlled
+ * whenever the edge proxy appends (as API Gateway does) rather than
+ * replaces. The last element is the edge proxy's own claim either way.
+ */
+const lastForwardedClientIp = (forwarded: string): string | undefined => {
+  const elements = forwarded.split(",")
+  for (let index = elements.length - 1; index >= 0; index--) {
+    const match = FORWARDED_FOR_CLIENT.exec(elements[index] ?? "")
+    if (match?.[1]) return match[1]
+  }
```
aliasunder: **[simplicity]** The reverse index loop with `elements[index] ?? ""` exists only to satisfy `noUncheckedIndexedAccess` — the bounds guarantee in-range access, so the fallback can never fire. `toReversed()` + `for...of` eliminates the index arithmetic, the bounds ceremony, and the type-checker fallback. Consistent with the codebase's existing `toSorted()` usage (same ES2023 family, same `lib: ES2025`).

```suggestion
  for (const element of forwarded.split(",").toReversed()) {
    const match = FORWARDED_FOR_CLIENT.exec(element)
    if (match?.[1]) return match[1]
  }
```

---
*🔍 ship-check · code-quality · claude-opus-4-6*

aliasunder (commented):
## Phase 2: Code Quality

1 finding across 16 files reviewed (11 production source/config, 5 markdown docs).

---
*🔍 ship-check · code-quality · claude-opus-4-6*

src/vault-mcp/oauth/__tests__/oauth-routes.test.ts
```diff
@@ -549,6 +555,91 @@ describe("OAuth endpoint rate limiting", () => {
...
+
+  it("logs the real peer IP — not the spoofed value — when the limit trips", async () => {
+    for (let i = 1; i <= 5; i++) {
+      await registerWithSpoofedIp("198.51.100.9")
+    }
+    logs.length = 0
+    const sixth = await registerWithSpoofedIp("198.51.100.9")
+    expect(sixth.status).toBe(429)
+    const event = logs.find((log) => log.message === "oauth_rate_limited")
+    expect(event).toMatchObject({
```
aliasunder, line 630: **[assertion]** `toMatchObject` omits `message: "oauth_rate_limited"` — the parallel trusted test at line 482 includes it in its `toMatchObject` alongside `level` and `data`. The `.find()` pre-filters by message so correctness isn’t at risk, but including `message` aligns with the established pattern and makes the assertion self-documenting if the `.find()` predicate drifts.

Suggested:
```ts
expect(event).toMatchObject({
  level: "warn",
  message: "oauth_rate_limited",
  data: expect.objectContaining({ path: "/register" }),
})
```

---
*🔍 ship-check · test-audit · claude-opus-4-6*

src/vault-mcp/server.ts
```diff
@@ -125,9 +130,11 @@ const startServer = async (): Promise<void> => {
...
 
   const app = express()
-  // Trust exactly one proxy hop (API Gateway). `true` would trust the entire
-  // X-Forwarded-For chain, letting clients spoof req.ip via injected headers.
-  app.set("trust proxy", 1)
+  // Proxy trust is deployment-explicit (TRUST_PROXY_HOPS): 0 in direct
+  // exposure, so an injected X-Forwarded-For can't spoof req.ip (the OAuth
+  // rate limiter's fallback bucket key). `true` would trust the entire
+  // chain, letting clients claim any IP via appended headers.
+  app.set("trust proxy", config.trustProxyHops)
```
aliasunder, line 140: **[coverage gap]** `TRUST_PROXY_HOPS` wiring has no integration test. Config parsing is covered (`config.test.ts` verifies default 0, valid integers, rejections), but no test proves this line actually changes Express’s `req.ip` resolution at runtime. A `TRUST_FORWARDED_HEADER=true` integration test exists and exercises the Forwarded-header path end-to-end, but the XFF/trust-proxy path has no equivalent.

Needs: an integration test (in `server-integration.test.ts`) that boots a server with `TRUST_PROXY_HOPS=1`, sends a request with an `X-Forwarded-For` header, and verifies the rate-limit bucket keys on the XFF-derived IP rather than the socket peer. This would catch a wiring regression (e.g. the config value not reaching `app.set`) that no current test detects.

---
*🔍 ship-check · test-audit · claude-opus-4-6*

aliasunder (commented):
## Phase 3: Test Audit

2 findings across 6 test files — 1 assertion quality, 1 coverage gap. All 6 changed test files audited per-`it()` against two-bar rule, assertion quality, hygiene, and completeness. Defensive guardrail tests are strong: the spoofed-header bypass, last-element bucket keying, and peer-IP log verification all clear both bars and genuinely prove the GHSA fix.

---
*🔍 ship-check · test-audit · claude-opus-4-6*

src/auth.ts
aliasunder: Fixed — applied `toReversed()` + `for...of` (with trailing-comment doc cleanup).

---
*🔍 ship-check · pr-monitor · openrouter/moonshotai/kimi-k3*

src/vault-mcp/oauth/__tests__/oauth-routes.test.ts
aliasunder: Fixed — added `message: "oauth_rate_limited"` to `toMatchObject`.

---
*🔍 ship-check · pr-monitor · openrouter/moonshotai/kimi-k3*

aliasunder:
## Triage follow-up: align the router test harness with the PR's trust model

Prescription for the flagged pre-existing gap from the Phase 1 review (`src/vault-mcp/mcp-core/__tests__/mcp-router.test.ts:169`), scoped into this PR. Current head (`b9ec93e`) doesn't touch this file, so line references hold.

**The gap.** The harness hardcodes `app.set("trust proxy", true)` while production (`server.ts`) now sets `app.set("trust proxy", config.trustProxyHops)` with default `0`. Every request carries `x-forwarded-for: 192.0.2.10` via `baseHeaders`, so under blanket trust `req.ip` is always `192.0.2.10` — the exact ambient-trust model this PR removed from production. 11 `clientIp: FORWARDED_IP` assertions depend on it (lines 601, 632, 639, 649, 683, 725, 756, 776, 796, 838, 889).

**Fix steps:**

1. **Derive trust from the config the router receives.** In `setupHarness`, hoist the existing fallback (`const config = opts.config ?? DEFAULT_CONFIG`) and use it in both places:
   ```ts
   app.set("trust proxy", config.trustProxyHops)   // was: app.set("trust proxy", true)
   ...
   config,                                          // line 176, same object
   ```
   `DEFAULT_CONFIG` is already `loadConfig({})`, so the harness default becomes production's default (`0`) for free.

2. **Pin the listener to IPv4** so the untrusted socket peer is deterministic: `app.listen(0, "127.0.0.1", ...)`. The current host-less `app.listen(0)` binds dual-stack, which renders the peer as the IPv4-mapped `::ffff:127.0.0.1` — pinning avoids encoding that form in assertions.

3. **Tests that assert an XFF-derived identity opt into trust explicitly.** Define one shared trusted-proxy config, e.g. `const TRUSTED_PROXY_CONFIG = loadConfig({ TRUST_PROXY_HOPS: "1" })`, and pass `config: TRUSTED_PROXY_CONFIG` in the tests that keep `clientIp: FORWARDED_IP`. Keeping the injected value (rather than flipping to loopback) is the stronger assertion — it proves the IP comes from the request, not a constant — and the trust is now per-test and visible instead of ambient.

4. **Flip the GHSA regression test (line 673) to default config + socket peer.** With `trustProxyHops: 0`, assert `clientIp: "127.0.0.1"`. This strengthens the test: it now proves that under production defaults *neither* the Forwarded header *nor* a client-supplied X-Forwarded-For shifts the logged identity. Update its comment accordingly — "the XFF-derived req.ip is the log identity" stops being accurate once XFF is untrusted; the identity is the socket peer.

5. **No change needed** to the trusted-Forwarded test (line 654) — `extractClientIp` short-circuits on the Forwarded header before `req.ip` is consulted.

Counted lift: 1 harness line + 1 listen host + 11 assertion sites (shared config constant covers most) + 1 comment update, all in one file.

---
*🔍 ship-check · triage · claude-fable-5*

docker-compose.yml
aliasunder: Fixed at 248c203 — both `deploy.yml` and `test_deploy.yml` now take `TRUST_PROXY_HOPS` / `TRUST_FORWARDED_HEADER` from repo Variables and conditionally write them to the Lightsail `.env` (unset Variable = no line written, compose defaults apply).

---

src/vault-mcp/server.ts
aliasunder: Fixed at d63a159 — added a `TRUST_PROXY_HOPS=1` integration describe in `server-integration.test.ts`: five requests with the same `X-Forwarded-For` exhaust one bucket (201×5 → 429) while a different XFF value gets a fresh bucket (201). The fresh-bucket assertion is the wiring detector — if the config value never reached `app.set("trust proxy", …)`, every request would share the socket peer's bucket and the second client would 429.

---

deploy/remote/README.md
```diff
@@ -466,6 +470,7 @@ required. These optional settings are worth knowing about:
 | `SYNC_CONFIGS`       | daily notes + plugin settings | Obsidian settings categories synced to the server (see [Daily notes](#daily-notes)); `none` disables                                   |
 | `DAILY_NOTES_FOLDER` | from vault config             | Sets the daily notes folder (see [Daily notes](#daily-notes))                                                                          |
 | `DAILY_NOTES_FORMAT` | from vault config             | Sets the daily note filename format (see [Daily notes](#daily-notes))                                                                  |
+| `TRUST_PROXY_HOPS`   | `0`                           | Set `1` when a tunnel or reverse proxy fronts the server — OAuth rate limiting then buckets by the real client IP, not the proxy's     |
```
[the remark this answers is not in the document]
aliasunder: Fixed at d63a159 — added the `TRUST_FORWARDED_HEADER` row (default `false`, set `true` when the fronting proxy sets the RFC 7239 `Forwarded` header, e.g. AWS API Gateway) adjacent to the `TRUST_PROXY_HOPS` row.

---

src/auth.ts
```diff
@@ -30,19 +30,37 @@ export const parseBearer = (header: string | undefined): string | null => {
...
 /**
- * Real client IP for logging and rate limiting. API Gateway sends it in
- * the RFC 7239 Forwarded header, which Express never reads — behind the
- * gateway, req.ip is the gateway's own egress node. The first `for=`
- * element is the original client (proxies append); req.ip is the
- * fallback for proxies that don't send Forwarded.
+ * The last `for=` value in an RFC 7239 Forwarded header — the claim of the
+ * proxy closest to the server. Walking right-to-left matters: a client can
+ * whenever the edge proxy appends (as API Gateway does) rather than
+ * replaces. The last element is the edge proxy's own claim either way.
```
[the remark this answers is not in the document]
aliasunder: Fixed at d63a159 — restored the missing verb phrase: "a client can prepend its own elements whenever the edge proxy appends (as API Gateway does) rather than replaces — the last element is the edge proxy's own claim either way."

---

aliasunder:
Harness prescription from [the triage comment](https://github.com/aliasunder/vault-cortex/pull/468#issuecomment-5351406243) implemented at d63a159 — all five steps: config-derived `app.set("trust proxy", …)`, IPv4-pinned listener, shared `TRUSTED_PROXY_CONFIG` opt-in for the 10 XFF-identity assertions, GHSA regression test flipped to production defaults asserting the socket peer, trusted-Forwarded test untouched. Mutation-verified: reverting the harness to blanket trust fails exactly the GHSA test.

---

aliasunder:
Both PR-level findings above addressed at 6388f44:

- **DEPLOY.md CI/CD Variables table** — added `TRUST_PROXY_HOPS` and `TRUST_FORWARDED_HEADER` rows with the Lightsail semantics (compose defaults the instance to `1`/`true` for API Gateway; `ORIGIN_URL` deployments set `2`/`false` so the `X-Forwarded-For` chain carries the client).
- **server.json environmentVariables** — added both vars to the manifest with server defaults (`0`/`false`), mirroring the README config table.

---

src/vault-mcp/config.ts
```diff
@@ -223,6 +234,24 @@ export const loadConfig = (
...
   )
 
+  // Default 0 = direct exposure: req.ip is the socket peer and injected XFF
+  // can't shift a client into a fresh rate-limit bucket. A proxy-fronted
+  // deployment opts into one hop per hop it controls.
+  const trustProxyHops = envVar
+    .from(env)
+    .get("TRUST_PROXY_HOPS")
+    .default("0")
+    .asIntPositive()
```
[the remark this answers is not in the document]
aliasunder: The described mechanism does not occur — verified empirically against Express. With `trust proxy = 2` and a single-entry `X-Forwarded-For: 203.0.113.5`, `req.ip` is `203.0.113.5`, not the socket peer: when the hop count meets or exceeds the chain length, proxy-addr yields the furthest (leftmost) entry, so distinct clients still key distinct buckets and the described egress-IP collapse cannot happen. The genuine overcount hazard runs the opposite direction — the leftmost entry is client-forgeable when the count exceeds the machines actually in the chain — but that is a topology misconfiguration the server cannot validate at boot (it cannot know how many proxies exist), an arbitrary numeric cap like 4 would not prevent it (the failing example is hops=2 with one proxy), and a warn-whenever-set would fire on every correctly configured deployment. The per-topology guidance (README/DOCKERHUB tables, deploy READMEs, DEPLOY.md, .env templates) is the mitigation: count the proxies you control.

---

src/__tests__/integration/server-integration.test.ts
```diff
@@ -621,6 +621,206 @@ describe("default config", () => {
...
+    // share the socket peer's bucket.
+    it("spoofed Forwarded headers do not bypass the /register rate limit", async () => {
+      for (let i = 1; i <= 5; i++) {
+        const response = await register(`198.51.100.${i}`)
+        expect(response.status).toBe(201)
+      }
+      const sixth = await register("198.51.100.6")
+      expect(sixth.status).toBe(429)
+    })
+  })
```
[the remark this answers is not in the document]
aliasunder: This test already exists — the top-level describe `X-Forwarded-For rate limiting (default proxy trust)` (lines 656–704 of this file, added in 248c203 for the earlier round of this same finding) boots a server with no env overrides (default config, `TRUST_PROXY_HOPS` unset → 0), sends five `/register` requests each with a distinct `X-Forwarded-For` value and no `Forwarded` header, and asserts the sixth returns 429. It sits in its own describe rather than inside the default-config `OAuth rate limiting` block because the rate limiter's in-memory store is per-process: the Forwarded test there leaves the shared 5-req bucket exhausted inside the 60s window, so this test needs a dedicated default-config server — the comment at the top of the describe documents exactly that. A regression re-raising the hop count fails this test.

---

PR #473, opened 2026-08-20
chore: bump umm-actually to v0.3.11

aliasunder:
Pins `umm_review.yml` to umm-actually v0.3.11 (`0422383`). Every posted finding comment now ends with an `umm-actually · <model>` byline naming the model that produced it.

### DIFF
--- a/.github/workflows/umm_review.yml
+++ b/.github/workflows/umm_review.yml
@@ -77,7 +77,7 @@ jobs:
       # to the action's defaults. Inputs whose default is empty
       # (fallback_model, max_findings) pass empty when the var is unset —
       # identical to omitting them.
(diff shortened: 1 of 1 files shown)


aliasunder:
Re: umm-actually **Grant pull-requests: read to the GITHUB_TOKEN used by the gh api step** — fixed in e9e3d50: `pull-requests: read` added to both the workflow-level and job-level permissions blocks. No run has ever exercised the comment-trigger path (every `issue_comment` run on record is guard-skipped), so the failure is unconfirmed, but the grant is what the pulls endpoint documents and it adds nothing beyond what the step already assumes. Swept the other workflows: the release workflows already pair their PR reads with `pull-requests: read`; this was the only gap.

---

PR #489, opened 2026-08-24
chore: replace CLAUDE.md with symlink to AGENTS.md

aliasunder:
## Summary
- Replace CLAUDE.md (which was just `@AGENTS.md` + an empty tips section) with a symlink to AGENTS.md
- Claude Code auto-reads CLAUDE.md and follows the symlink — same content, no indirection

## Test plan
- [ ] Verify `ls -la CLAUDE.md` shows `CLAUDE.md -> AGENTS.md`
- [ ] Verify Claude Code reads AGENTS.md content when auto-loading CLAUDE.md

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,7 +1,1 @@
-# CLAUDE.md
-
-@AGENTS.md
-
-## Claude-Specific Tips
-
-<!-- Any Claude Code specific tips go here -->
+AGENTS.md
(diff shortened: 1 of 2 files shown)


aliasunder:
Addressed in bdc7787: added a Windows `core.symlinks=true` note to CONTRIBUTING.md. The symlink is intentional — it eliminates the `@AGENTS.md` import indirection so Claude Code reads AGENTS.md directly. The portability trade-off is documented rather than avoided, since the symlink serves every checkout that resolves them (macOS, Linux, Windows with the setting enabled).

PR #506, opened 2026-08-29
repro(deploy): drop authorizer identity source to reproduce the 403 on tokenless /mcp

aliasunder:
Test-deploy only. Never merge.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/sst.config.ts
+++ b/sst.config.ts
@@ -339,7 +339,8 @@ export default $config({
           timeout: "5 seconds",
           memory: "128 MB",
         },
-        identitySources: ["$request.header.Authorization"],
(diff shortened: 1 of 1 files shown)


aliasunder:
Reproduction branch only. Removing the authorizer's identity source (`identitySources: []`) put the gateway back in the pre-fix state: a tokenless `POST /mcp` returned `403 {"message":"Forbidden"}` on a test deploy while the discovery routes stayed 200, which confirmed the behaviour described in #121. The 401 wiring has been redeployed and verified. Closing without merge.

PR #507, opened 2026-08-30
fix(oauth): grant the vault scope when a client requests none

aliasunder:
## What

- `authorize` in `oauth-provider.ts` defaults the requested scope to `vault` when the client sends none or a blank `scope=`. The default is applied once, before the request is stored, so the consent page, the code exchange, and every refresh read the same value.
- `DEFAULT_SCOPE` (`"vault" as const`) names the server's one scope; the static token's `AuthInfo` and the metadata's `scopes_supported` in `oauth-routes.ts` both read it.
- ARCHITECTURE.md: one sentence under the JWT payload describing the rule.

## Why

A client that requested no scope got an access token with an empty `scope` claim and a consent page reading "No specific scopes requested", while the static token carries `vault` and the metadata advertises `vault` as the only supported scope. Nothing reads the claim today, so this is cosmetic, but it becomes a real seam once tokens carry more claims and layers start comparing them.

## Behaviour

| Request | Before | After |
| --- | --- | --- |
| no `scope` | consent: "No specific scopes requested"; token `scope: ""` | consent lists `vault`; token `scope: "vault"` |
| `scope=` (blank) | consent: empty list item; token `scope: ""` | same as above |
| `scope=vault` | unchanged | unchanged |
| `scope=read` | `read` | `read` (the default never widens a named scope) |

`scopes_supported` is unchanged; a refresh still may narrow but never widen the grant.

## Tests

- Five provider tests: three grant `vault` (no scope `[]`, blank scope `[""]`, `scopes` key absent) and two keep an explicit scope (`vault`, narrower `read`). Each asserts the `oauth_authorize_started` log, the rendered consent list, and the issued token's `scope`; the no-scope case also asserts the verified `AuthInfo.scopes` and the refreshed token's scope.
- Mutation check: with the default removed, the three "grants vault" tests fail and the two explicit-scope tests still pass.
- Full suite: 3,294 tests green.

No automated per-PR coverage over HTTP for the no-scope case; the route tests exercise `scope=vault` end to end and the provider tests cover the default at the boundary where the SDK hands over `scopes`.

### DIFF
--- a/src/vault-mcp/oauth/__tests__/oauth-provider.test.ts
+++ b/src/vault-mcp/oauth/__tests__/oauth-provider.test.ts
@@ -839,6 +839,129 @@ const startAuthFlow = async (
   return match[1]
 }
 
+describe("OAuth default scope", () => {
+  /** Starts an authorization flow with the given scopes and returns the
+   *  rendered consent HTML plus the request id it carries. */
+  const startAuthFlowWithScopes = async (
+    oauth: OAuthProvider,
+    client: OAuthClientInformationFull,
+    scopes: string[] | undefined,
+  ): Promise<{ consentHtml: string; requestId: string }> => {
+    let consentHtml = ""
+    const res = {
+      type: () => res,
+      send: (html: string) => {
+        consentHtml = html
+        return res
+      },
+    }
+    await oauth.provider.authorize(
+      client,
+      {
+        codeChallenge: "test-challenge",
+        redirectUri: "https://example.com/cb",
+        ...(scopes === undefined ? {} : { scopes }),
+      },
+      res as never,
+    )
+    const requestId = /name="request_id"\s+value="([^"]+)"/.exec(
+      consentHtml,
+    )?.[1]
+    if (!requestId) throw new Error("no request_id in consent HTML")
+    return { consentHtml, requestId }
+  }
+
+  it("grants vault when the request names no scope", async () => {
+    const { logs, testLogger, oauth, client } = await setupAuditTest()
+
+    const { consentHtml, requestId } = await startAuthFlowWithScopes(
+      oauth,
+      client,
+      [],
+    )
+
+    const started = logs.find(
+      (log) => log.message === "oauth_authorize_started",
+    )
+    expect(started?.data.scopes).toEqual(["vault"])
+    expect(consentHtml).toContain("<li>vault</li>")
+    expect(consentHtml).not.toContain("No specific scopes requested")
+
+    const code = oauth.approveRequest(requestId, testLogger)
+    const issued = await oauth.provider.exchangeAuthorizationCode(client, code)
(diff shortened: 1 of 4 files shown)


aliasunder:
The umm-actually finding above is a placeholder: title, location (`...:1`), failure scenario, and suggested fix are all literal `...`, so there is nothing to evaluate. The bot re-reviewed at fb3005c with no new findings. Every CI check is green and there are no open review threads.

---

src/vault-mcp/oauth/oauth-provider.ts
```diff
@@ -45,6 +45,10 @@ const ACCESS_TOKEN_TTL_S = 24 * 60 * 60
 const REFRESH_TOKEN_TTL_S = 60 * 24 * 60 * 60
 // 10 minutes — OAuth spec recommends short auth code lifetimes.
 const AUTH_CODE_TTL_S = 10 * 60
+// The one scope the server grants; also what the static token carries
+// (see verifyAccessToken) and what the metadata advertises as
+// scopes_supported.
+export const DEFAULT_SCOPE = "vault"
```
aliasunder: Let's make this "as const"

aliasunder: Done in 0b5fcb9: `export const DEFAULT_SCOPE = "vault" as const`. Lint, tsc, and the OAuth tests are clean.

---
*Reply drafted by Claude Code (ship-check pr-monitor)*

src/vault-mcp/oauth/__tests__/oauth-provider.test.ts
```diff
@@ -839,6 +839,151 @@ const startAuthFlow = async (
...
+      (log) => log.message === "oauth_authorize_started",
+    )
+    expect(started?.data.scopes).toEqual(["vault"])
+    expect(consentHtml).toContain("<li>vault</li>")
+
+    const code = oauth.approveRequest(requestId, testLogger)
+    const issued = await oauth.provider.exchangeAuthorizationCode(client, code)
+    expect(issued.scope).toBe("vault")
+  })
+})
```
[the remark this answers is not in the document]
aliasunder: Correct: the test audit added a fifth case (scopes key absent) after the body was written. The Tests section now says five provider tests, three granting vault (no scope, blank scope, scopes key absent) and two keeping an explicit scope, with the mutation result updated to three failing; the What section also names the scopes_supported change in oauth-routes.ts.

---
*Reply drafted by Claude Code (ship-check pr-monitor)*

aliasunder:
Live check on the test deploy of 7ad0d7e (PR #508, which includes this change): an authorization request with no `scope` renders the consent list as `["vault"]` with no "No specific scopes requested" fallback, the code exchange returns `scope: vault` (also in the JWT claim), and a refresh carries `vault` through.

---

PR #520, opened 2026-09-02
feat(setup): name the platform's settings tab on the setup page

aliasunder:
## What

The `/setup` page now names the platform's settings tab instead of "your deployment's settings" when it can tell which container hosting platform it runs on:

| Platform | Detected by | Phrase |
| --- | --- | --- |
| Render | `RENDER_EXTERNAL_URL` | the service's Environment tab on Render |
| Railway | `RAILWAY_PUBLIC_DOMAIN` | the service's Variables tab on Railway |
| anything else | neither set | your deployment's settings (unchanged) |

The phrase appears wherever the page tells the owner to find or fix a variable: the `MCP_AUTH_TOKEN` hint under the token field, the wrong-token error, and the four "One more setting" remedies (`VAULT_NAME` unset or not found, `VAULT_PASSWORD` missing or rejected). The wording matches the Railway and Render guides word for word, so the page and the guide a user is following agree.

## How

- `setup-page.ts`: `HostingPlatform` type, `settingsLocation()` helper, an optional `hostingPlatform` on the `sign-in` and `blocked` views; the six sentences compose with the phrase.
- `setup-server.ts`: `detectHostingPlatform()` reads the two platform variables via `env-var`, Render first, the same order as `print-derived-env`'s `PUBLIC_URL` derivation so the page never names a different platform than the one the URL came from. Detection does not depend on whether `PUBLIC_URL` was set by hand.
- `setup-routes.ts`: `hostingPlatform` option, threaded into the sign-in and blocked views and the 401 message.
- `ARCHITECTURE.md`: one sentence in the setup-mode list.

The full server's "Already set up" page is untouched.

## Tests

| Tier | Covers | Result |
| --- | --- | --- |
| `setup-page.test.ts` (+6) | exact hint and remedy sentences per platform; the generic sentences pinned verbatim | pass |
| `setup-routes.test.ts` (+3) | the platform reaches the sign-in view, the 401 error, and the blocked view | pass |
| `setup-server.test.ts` (+5) | real entry point with a controlled env: Render, Railway, both set (Render wins), blank `RENDER_EXTERNAL_URL` treated as unset, neither | pass |
| `remote-image-boot.test.ts` (+1 assertion) | the platform variable reaches the setup server inside the `:remote` image through the s6 longrun | pass (51 tests, image built from this branch) |
| `npm test` | 3,506 pass; the one failure (`oauth-provider.test.ts` sliding-expiry) fails on `main` too and is unrelated: its expected value uses calendar days across a daylight-saving change | pre-existing |

Mutation checks (each reverted after): swapping the detection order failed the both-set test; dropping the blocked-view threading failed its routes test; rewording the generic phrase failed the pinned-generic tests. `npm run lint`, `npm run build`, and `knip` clean.

No live deploy: the copy is deterministic and the boot tier proves the environment inheritance.

## Summary by CodeRabbit

- **New Features**
  - Setup pages now provide platform-specific guidance for Render and Railway deployments.
  - Instructions identify the relevant Environment or Variables tab when users need to configure or fix settings.
  - Generic guidance remains available when the hosting platform cannot be detected.

- **Tests**
  - Added coverage for platform detection, platform-specific instructions, precedence, and fallback behavior.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -905,7 +905,10 @@ flowchart TD
    `/healthz` answers `{ ok: true, mode: "setup" }` within seconds (so a
    platform deploy goes live), `/setup` serves the sign-in page, browser
    GETs to any other path redirect to `/setup`, and API requests answer
-   503 with the setup URL.
+   503 with the setup URL. When `RENDER_EXTERNAL_URL` or
+   `RAILWAY_PUBLIC_DOMAIN` identifies the host, the page names that
+   platform's settings tab wherever it tells the owner to find or fix a
+   variable.
 4. `POST /setup` checks `MCP_AUTH_TOKEN` first, then signs in through
    Obsidian's account API (the same request `ob login` makes, two-factor
    included).
--- a/src/__tests__/docker/remote-image-boot.test.ts
+++ b/src/__tests__/docker/remote-image-boot.test.ts
@@ -1128,7 +1128,13 @@ describe("remote image boot — setup mode (no Sync token anywhere)", () => {
   it("serves the sign-in page and rejects a wrong MCP token without contacting Obsidian", async () => {
     const page = await fetch(`http://127.0.0.1:${port}/setup`)
     expect(page.status).toBe(200)
-    expect(await page.text()).toContain("<h1>Connect Obsidian Sync</h1>")
+    const html = await page.text()
+    expect(html).toContain("<h1>Connect Obsidian Sync</h1>")
+    // RAILWAY_PUBLIC_DOMAIN is a container env var, so the hint proves the
+    // s6 longrun hands the platform variable to the setup server.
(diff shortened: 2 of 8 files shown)


aliasunder:
Re umm-actually finding #5515233333 ("Name the platform's tab on the already-set-up page too", `setup-page.ts:275`): intentional. The "Already set up" page is served by the full server to any visitor without a login, while the setup page's platform hints sit behind the MCP token check and the Obsidian sign-in. Naming the hosting platform on the unauthenticated page would disclose more about the deployment than it does today, so the repo owner decided it keeps the generic wording rather than gaining platform detection in `server.ts`. Whether that page should be locked down is a separate decision, not part of this change.

---
*🔍 ship-check · pr-monitor · claude-fable-5-1*

aliasunder:
@coderabbit review

